PR #2, opened 2022-02-17
Parallel communication

ShaneRozenLevy:
This PR:

- Fixes issue with clock time
- adds a parallel option which is necessary for communicating with larger number of motors when the behavior takes longer
- Tested on Jerboa robot. Robot functions with both either flag


jdcaporale (approved):
LGTM

PR #3, opened 2022-03-04
Initial commit of Joint and robot class prototypes (including moteus …

jdcaporale:
…joints).

PR for comments only DO NOT MERGE


.gitignore
```diff
@@ -5,4 +5,5 @@ lcm_types/*.hpp
 lcm_types/*.py
 __pycache__/
 python_scripts/python_types/*
-python_scripts/logs/*
\ No newline at end of file
+python_scripts/logs/*
+.vscode*
```
ShaneRozenLevy, line 9: whats the logic behind this change?

include/kodlab_mjbots_sdk/jointBase.h
```diff
@@ -0,0 +1,84 @@
...
+ * @brief 
+ * Abstract Base class for any joint used in the robot class.
+ */
+class JointBase {
+    protected:
+        float gear_ratio_ = 1.0;
+        float zero_offset_ = 0;
+        int  direction_ = 1; 
+        float max_torque_ = std::numeric_limits<float>::infinity();
+        float pos_limit_min_ = -std::numeric_limits<float>::infinity();
```
ShaneRozenLevy, line 35: Will need to describe what these actually mean. What happens if you exceed that limit?

include/kodlab_mjbots_sdk/jointBase.h
```diff
@@ -0,0 +1,84 @@
...
+        float pos_limit_max_ = -std::numeric_limits<float>::infinity();
+        bool  soft_stop_active = true;
```
ShaneRozenLevy, line 37: same with this one

include/kodlab_mjbots_sdk/jointMoteus.h
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * @brief 
+     * 
+     */
+    public:
+        const int can_bus;
+        const int can_id;
+
+        mjbots::moteus::Mode mode_ = ::mjbots::moteus::Mode::kStopped;
+        mjbots::moteus::Pi3HatMoteusInterface::ServoCommand command;
```
ShaneRozenLevy, line 19: I'd recommend that on your first pass through you don't mess around with the structs in the next 4 lines of code since they can be confusing and cause bugs if not handled properly.

include/kodlab_mjbots_sdk/jointMoteus.h
```diff
@@ -0,0 +1,66 @@
...
+                    res.velocity = ::mjbots::moteus::Resolution::kIgnore;
+                    res.feedforward_torque = ::mjbots::moteus::Resolution::kInt16;
+                    res.kp_scale = ::mjbots::moteus::Resolution::kIgnore;
+                    res.kd_scale = ::mjbots::moteus::Resolution::kIgnore;
+                    res.maximum_torque = ::mjbots::moteus::Resolution::kIgnore;
+                    res.stop_position = ::mjbots::moteus::Resolution::kIgnore;
+                    res.watchdog_timeout = ::mjbots::moteus::Resolution::kIgnore;
+
+                    // Minimal query reads for speed 
+                    query.mode = Resolution::kInt8;
```
ShaneRozenLevy, line 45: Currently these are set in the moteusProtocol.h likely you would need to modify the moteusProtocol to make it use this

include/kodlab_mjbots_sdk/jointMoteus.h
```diff
@@ -0,0 +1,66 @@
...
+                    query.torque = Resolution::kIgnore;
+                    query.q_current = Resolution::kIgnore;
+                    query.d_current = Resolution::kIgnore;
+                    query.rezero_state = Resolution::kIgnore;
+                    query.voltage = Resolution::kIgnore;
+                    query.temperature = Resolution::kIgnore;
+                    query.fault = Resolution::kIgnore;
+        }
+        
+        void updateMoteus(::mjbots::moteus::QueryResult servo_reply){
```
ShaneRozenLevy, line 57: Should probably add a static function Get to this class

include/kodlab_mjbots_sdk/jointMoteus.h
```diff
@@ -0,0 +1,66 @@
...
+            reply_ = servo_reply;
+            updateMoteus(servo_reply.position,servo_reply.velocity, servo_reply.mode);
+        }
+        void updateMoteus(float reply_pos, float reply_vel, ::mjbots::moteus::Mode mode ){
```
ShaneRozenLevy, line 61: This function should probably not exist in the child and instead live as a version in the parent.

include/kodlab_mjbots_sdk/jointMoteus.h
```diff
@@ -0,0 +1,66 @@
...
+            position_ = direction * (raw_pos * 2 * M_PI) + zero_offset;
```
ShaneRozenLevy, line 62: You ignore gear ratio here

include/kodlab_mjbots_sdk/jointMoteus.h
```diff
@@ -0,0 +1,66 @@
...
+     * 
+     */
+    public:
+        const int can_bus;
+        const int can_id;
+
+        mjbots::moteus::Mode mode_ = ::mjbots::moteus::Mode::kStopped;
+        mjbots::moteus::Pi3HatMoteusInterface::ServoCommand command;
+        mjbots::moteus::PositionResolution res;
+        mjbots::moteus::QueryCommand query;
```
ShaneRozenLevy, line 21: You don't appear to be properly setting up the query

include/kodlab_mjbots_sdk/jointMoteus.h
```diff
@@ -0,0 +1,66 @@
...
+                    query.q_current = Resolution::kIgnore;
+                    query.d_current = Resolution::kIgnore;
+                    query.rezero_state = Resolution::kIgnore;
+                    query.voltage = Resolution::kIgnore;
+                    query.temperature = Resolution::kIgnore;
+                    query.fault = Resolution::kIgnore;
+        }
+        
+        void updateMoteus(::mjbots::moteus::QueryResult servo_reply){
+            reply_ = servo_reply;
```
ShaneRozenLevy, line 58: Why copy the result?

ShaneRozenLevy (changes requested):
You are changing too much at one time. The idea is correct, but you are changing too much. Start with the simplest possible joint class that works with the moteus and make that work.

include/kodlab_mjbots_sdk/jointMoteus.h
jdcaporale: Because moteus has it built in to its position output in the base class the gearratio will be included

include/kodlab_mjbots_sdk/jointMoteus.h
jdcaporale: honestly no reason, this is one of those places where I dont know enough about the reply to know if you ever use it again or if we would want to

include/kodlab_mjbots_sdk/jointMoteus.h
jdcaporale: There was a simpler version in the parent, but it didnt include mode. and I wasnt sure how kosher it was to override withdifferent arguments. I guess if I just overload it it should be fine?

include/kodlab_mjbots_sdk/jointMoteus.h
jdcaporale: what would a static get be for exactly?

include/kodlab_mjbots_sdk/jointMoteus.h
jdcaporale: My hope was not to do that  and set them to something other than their defaults in teh constructor.

include/kodlab_mjbots_sdk/jointBase.h
jdcaporale: yes agreed wanted to run by you what the plan would be for these. also the defaults are describe in too many places maybe

.gitignore
jdcaporale: this one likely shouldnt have been pushed. 
their are vscode files I didnt want to see

include/kodlab_mjbots_sdk/jointMoteus.h
ShaneRozenLevy: look at the current function `Get` that I have in the `MjbotsRobotInterface`

include/kodlab_mjbots_sdk/jointMoteus.h
ShaneRozenLevy: Or have this one call the child one after handling the mode.

include/kodlab_mjbots_sdk/jointMoteus.h
ShaneRozenLevy: Yet you allow the user to set the gear ratio. Its horrible design to allow a user to set a variable that does nothing. Either we should support also setting a gear ratio hear or not allow it to be set for the moteus.

PR #4, opened 2022-03-10
VoidLCM and Pi3hat firmware

jdcaporale:
Changing the voidlcm implementation to work with some of the default functions we are using. Also adjusting pi3hat.h to work with the updated pi3hat firmware


ShaneRozenLevy:
LGTM,but name is bad.

PR #8, opened 2022-04-29
Added catch to robot interface

ShaneRozenLevy:
Adds a check to make sure the can packet actually arrived for all servos. If a servo is missing, then the state will not get updated. This mainly serves as a temporary fix and a method to observe if the packet issues continue.


jdcaporale (approved):

jdcaporale:
LGTM

PR #9, opened 2022-05-11
Bug in ordering of the moteus joint initialization

jdcaporale:
Fixing the ordering of the inputs in moteus joint class

### DIFF
--- a/include/kodlab_mjbots_sdk/joint_moteus.h
+++ b/include/kodlab_mjbots_sdk/joint_moteus.h
@@ -42,8 +42,8 @@ class JointMoteus: public JointBase{
             int can_bus,
             int direction = 1, 
             float zero_offset = 0,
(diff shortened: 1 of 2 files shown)


include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -42,8 +42,8 @@ class JointMoteus: public JointBase{
             int can_bus,
             int direction = 1, 
             float zero_offset = 0,
-            float max_torque = std::numeric_limits<float>::infinity(),
             float gear_ratio = 1.0, 
+            float max_torque = std::numeric_limits<float>::infinity(),
```
ShaneRozenLevy, line 46: Do we need to change any of the examples as well?

ShaneRozenLevy (approved):

PR #10, opened 2022-05-31
Imu experiments

ShaneRozenLevy:
- Added read of imu on pi3hat at the robot and control loop level
- Enabled setting imu update frequency and mounting configuration
- I also had to fix the bug with the soft start max torque not being used in the robot contructor

### DIFF
--- a/include/kodlab_mjbots_sdk/mjbots_robot_interface.h
+++ b/include/kodlab_mjbots_sdk/mjbots_robot_interface.h
@@ -37,11 +37,15 @@ class MjbotsRobotInterface {
    * @param realtime_params the realtime parameters defining cpu and realtime priority
    * @param soft_start_duration how long in dt to spend ramping the torque
    * @param robot_max_torque the maximum torque to allow per motor in the robot
+   * @param imu_mounting_deg Orientation of the imu on the pi3hat. Assumes gravity points in the +z direction
+   * @param imu_rate_hz Frequency of the imu updates from the pi3hat
    */
   MjbotsRobotInterface(const std::vector<JointMoteus> &joint_list,
                        const RealtimeParams &realtime_params,
                        int soft_start_duration = 1,
-                       float robot_max_torque = 100);
+                       float robot_max_torque = 100,
+                       ::mjbots::pi3hat::Euler imu_mounting_deg = ::mjbots::pi3hat::Euler(),
+                       int imu_rate_hz = 1000);
 
   /**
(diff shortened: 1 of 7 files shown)


include/kodlab_mjbots_sdk/mjbots_robot_interface.h
```diff
@@ -124,6 +134,7 @@ class MjbotsRobotInterface {
   std::future<::mjbots::moteus::Pi3HatMoteusInterface::Output> can_result_;      /// future can result, used to check if
   /// response is ready
   SoftStart soft_start_;                                                      /// Soft Start object
+  ::mjbots::pi3hat::Attitude attitude_;                                       /// Robot attitude
```
jdcaporale, line 137: I think if you are going to use smart pointers, they should be smart pointers everywhere including in the owner. Otherwise we are defeating the point of them.

jdcaporale (commented):
Likely needs some documentation, but looks good other than the smart pointer missing. 
Ethan has built out a class version of what you did based on this branch, I don't know if we want to push this first, or just wait til his is ready in the next couple days.

include/kodlab_mjbots_sdk/pi3hat_moteus_interface.h
```diff
@@ -203,7 +211,7 @@ class Pi3HatMoteusInterface : public kodlab::AbstractRealtimeObject{
     input.tx_can = { tx_can_.data(), tx_can_.size() };
     input.rx_can = { rx_can_.data(), rx_can_.size() };
     input.request_attitude = true;
-    input.attitude = &attitude_;
+    input.attitude = attitude_.get();
```
ShaneRozenLevy, line 214: @jdcaporale  So right now nothing at the pi3hat level uses smart pointers. I think its fine to keep it that way, or it should be a separate PR

include/kodlab_mjbots_sdk/mjbots_robot_interface.h
ShaneRozenLevy: So its important that this is a copy of the pi3hat::Attitude object that is passed into the pi3hat moteus interface other wise you get a race condition. For this object I don't ever use it as a smart ptr.

ShaneRozenLevy:
I'm not in a spot to finish up this PR, so I'm happy with either option in terms of getting this across the finish line.

In terms of what makes an easy to review PR I would recommend merging this one in, then Ethan's to keep everything small.

PR #13, opened 2022-06-15
adds live plotter python script  

jdcaporale:
Add a live plotter example to the kodlab_mjbots_sdk. Not perfect (the plot takes time and can't read LCM in that time so there are drops) but useful functionality

### DIFF
--- a/python_scripts/live_motor_plotter.py
+++ b/python_scripts/live_motor_plotter.py
@@ -1,5 +1,6 @@
 from xml.etree.ElementTree import PI
 import lcm
+from matplotlib import animation
 from lcm_types.ManyMotorLog import ManyMotorLog
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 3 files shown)


python_scripts/live_motor_plotter.py
```diff
@@ -0,0 +1,227 @@
...
+# Queues
+timestamps = collections.deque(np.zeros(WINDOWLENGTH))
+margins = collections.deque(np.zeros(WINDOWLENGTH))
+message_duration = collections.deque(np.zeros(WINDOWLENGTH))
+positions = collections.deque(np.zeros((WINDOWLENGTH,13)))
+velocities = collections.deque(np.zeros((WINDOWLENGTH,13)))
+torques = collections.deque(np.zeros((WINDOWLENGTH,13)))
+velo_filt = collections.deque(np.zeros((WINDOWLENGTH,13)))
+
+class BlitManager:
```
ethanmusser, line 30: Class template pulled from https://matplotlib.org/stable/tutorials/advanced/blitting.html

ethanmusser (approved):

ShaneRozenLevy (changes requested):
To make this really useful you should make it an abstract class that can be inherited, that way if you want to make multiple live plotting scripts you don't have to copy code. Beyond that LGTM.

jdcaporale:
@ShaneRozenLevy  ok if I merge in as is and add an issue? Agree it would be more useful, but in case people need some live plotting now I feel like it should be available.

ShaneRozenLevy:
Looking more closely, the `BlitManager` is general purpose. Just put it in a different file and it lgtm

PR #14, opened 2022-06-27
updates cmake for raspberry-pi4 and adds optimizations

jdcaporale:
This change sets up the "correct" flags for the raspberry pi's and adds the "-O3" optimization flag. 

There is some ambiguity in the use of a 32 bit OS in 64 bit hardware, but I imagine that given the tool chain and the fact that it is running the system figures that out under the hood.

### DIFF
--- a/cmake/pi.cmake
+++ b/cmake/pi.cmake
@@ -26,14 +26,16 @@
 # https://github.com/Pro/raspi-toolchain/blob/master/Toolchain-rpi.cmake
 
 if("$ENV{RASPBERRY_VERSION}" STREQUAL "")
+	message(WARNING "No Raspberry Pi version in environment, defaulting to Pi Version 1. May cause slow downs")
 	set(RASPBERRY_VERSION 1)
 else()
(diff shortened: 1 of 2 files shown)


cmake/pi.cmake
```diff
@@ -26,14 +26,16 @@
 # https://github.com/Pro/raspi-toolchain/blob/master/Toolchain-rpi.cmake
 
 if("$ENV{RASPBERRY_VERSION}" STREQUAL "")
```
ShaneRozenLevy, line 28: Is this something set in bashrc? If so update documentation to include that. Otherwise I would recommend maybe having someone else test this to make sure it does not break things, but LGTM. Thanks for putting in the work on this.

ShaneRozenLevy (approved):

cmake/pi.cmake
jdcaporale: I was thinking the same thing. I tested it on our stuff, but not much in just kodlab world. I will do before merging. This is set in the BASHRC and it is defaulted to 1. I will add a change to the documentation.

PR #15, opened 2022-07-06
Joint smart pointers

jdcaporale:
Previously the joints were passed in by the main function through the control loop and copied as members into the robot interface. Now the main function "std::move"s  the joints (removing a possibly confusing copy). Within the robot interface the joints are now kept as shared pointers to all allow for future integration into higher level limb and robot kinematic/dynamics classes. This vector of shared pointers to joints is changed to a public member for easier access.
Finally, getter functions are added for getting a set of joint via indices.

### DIFF
--- a/include/kodlab_mjbots_sdk/mjbots_robot_interface.h
+++ b/include/kodlab_mjbots_sdk/mjbots_robot_interface.h
@@ -108,11 +108,12 @@ class MjbotsRobotInterface {
   int num_servos_;                         /// The number of motors in the robot
   std::map<int, int> servo_bus_map_;       /// map from servo id to servo bus
 
-  std::vector<JointMoteus> joints_; /// Joint vector for the robot, owns all state information
+  // std::vector<JointMoteus> joints_; /// Joint vector for the robot, owns all state information
+  std::vector< std::shared_ptr<JointMoteus>> joints_;
   std::vector<std::reference_wrapper<const float>> positions_;  /// Vector of the motor positions (references to the members of joints_)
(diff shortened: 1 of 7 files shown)


src/mjbots_robot_interface.cpp
```diff
@@ -137,6 +144,25 @@ std::vector<::mjbots::moteus::Mode> MjbotsRobotInterface::GetJointModes() {
...
 
+std::vector<std::shared_ptr<::kodlab::mjbots::JointMoteus>> MjbotsRobotInterface::GetJoints(std::vector<int> joint_indices){
+  std::vector<std::shared_ptr<::kodlab::mjbots::JointMoteus>> joint_list;
+  for (int ind: joint_indices){
+    joint_list.emplace_back(joints[ind]);
+  }
+  return joint_list;
+}
+
+std::vector<std::shared_ptr<::kodlab::mjbots::JointMoteus>> MjbotsRobotInterface::GetJoints(std::initializer_list<int> joint_indices){
```
ShaneRozenLevy, line 155: what is the purpose of this function and the next?

include/kodlab_mjbots_sdk/mjbots_robot_interface.h
```diff
@@ -104,15 +108,32 @@ class MjbotsRobotInterface {
    */
   std::vector<::mjbots::moteus::Mode> GetJointModes();
 
+
+  /**
+   * @brief Get vector of shared_ptr to joint objects
+   * 
+   * @param joint_indices
```
ShaneRozenLevy, line 115: Add comment to in and out describing what the input and output are.

ShaneRozenLevy (approved):
A minor request for documentation, but otherwise looks good.

Issue #11, opened 2022-05-31
Create Attitude Class


ShaneRozenLevy:
Currently the attitude class lives in the pi3hat. It should really live in its own file with its useful functions.

ethanmusser:
Check out the approach in #16.

PR #16, opened 2022-07-11
Implementation of Attitude Class & Associated Math Headers

ethanmusser:
Implements an `Attitude` class in the `kodlab` namespace, which stores IMU data and provides methods for converting between rotation representations without repetitive computation.  Also provides headers with helper functions and objects to aid in rotation and general math calculations.

This PR builds on changes implemented in #10 and addresses #11. These changes are made to provide a structure for housing IMU data that can be expanded to additional IMU data, with convenient member functions and structures for direct usage in robot behaviors.

### Changes

- Provides an `Attitude` class in the `kodlab` namespace in `attitude.h`.  This class is a data structure for storing inertial measurements, with member functions for updating the internal data and converting between data representations without unnecessary recomputation.
- Replaces `moteus::pi3hat::Attitude` usages in `MjbotsRobotInterface` class with `kodlab::Attitude` shared pointers and corresponding accessors.
- Provides a sample logger for attitude data and an example `attitude_example.cpp` which demonstrates usage of the `Attitude` class.
- Provides the `rotations.h` header with structures and functions in the `kodlab::rotations` namespace.  This header contains a comprehensive `EulerAngles` object and several rotation conversion functions adapted from the [`RobotLocomotion/drake`](https://github.com/RobotLocomotion/drake/) repository (see function comments for details).  Together, these provide useful conversions and representations measuring and controlling attitude.
- Provides a `ValidatedCache` class in the `common_header.h` header.  This structure permits storage a generic type of data with a validity flag.  Valid data can be computed and cached, and recomputation is not required unless the data becomes invalid.
- Removes extraneous default rotation conversions in `pihat.h` and `pihat.cpp`.  These conversions should now be done via the `kodlab::Attitude` representation.
- Addresses a TODO in `pihat.h` and moves the relevant signum function implementation to a general `math.h` header file in the `kodlab::math` namespace.


include/kodlab_mjbots_sdk/common_header.h
```diff
@@ -31,4 +33,84 @@ static void EnableCtrlC() {
...
+
+  /**
+   * @brief Accessor for data (<b>does not check validity</b>)
+   * @return data
+   */
+  T get()
+  {
+    if (!valid_)
+    {
+      std::cerr << "[WARN] Returning invalid data." << std::endl;
```
jdcaporale, line 97: This mode of operation for get() where we allow but warn against invalid data is a design choice we made. I think we are happy with this since otherwise we would need to do something with null pointers or std::optional, but bringing it up jic.

include/kodlab_mjbots_sdk/common_header.h
```diff
@@ -3,9 +3,11 @@
 #pragma once
 #include <atomic>
 #include <signal.h>  // manage the ctrl+c signal
+#include <type_traits>
+#include <iostream>
 
 namespace kodlab {
-/**
+/*
```
jdcaporale, line 10: small comment error

src/CMakeLists.txt
```diff
@@ -13,15 +13,20 @@ set(HEADER_LIST "${kodlab_mjbots_sdk_SOURCE_DIR}/include/kodlab_mjbots_sdk/pi3ha
...
 
-set(SOURCE_LIST "pi3hat.cpp" 
+set(SOURCE_LIST "pi3hat.cpp"
                 "mjbots_robot_interface.cpp"
                 "polar_leg.cpp"
                 "soft_start.cpp"
                 "abstract_realtime_object.cpp"
                 "joint_base.cpp"
-                "cartesian_leg.cpp" )
+                "cartesian_leg.cpp"
```
jdcaporale, line 28: useful for future commits to not have to worry about the trailing parentheses

src/mjbots_robot_interface.cpp
```diff
@@ -181,7 +184,13 @@ void MjbotsRobotInterface::SetModeStop() {
 void MjbotsRobotInterface::Shutdown() {
   moteus_interface_->shutdown();
 }
-::mjbots::pi3hat::Attitude MjbotsRobotInterface::GetAttitude() {
-  return  attitude_;
+
+const ::kodlab::Attitude<float>& MjbotsRobotInterface::GetAttitude() {
+  return *attitude_;
```
jdcaporale, line 189: Could be in the header, but fine here too

examples/attitude_example.cpp
```diff
@@ -0,0 +1,150 @@
...
+      Eigen::AngleAxisf(M_PI, Eigen::Vector3f::UnitX()));
+    att_->set_world_offset(world_offset);
+    att_read_only_.set_world_offset(world_offset);
+  }
+
+private:
+  /**
+   * @brief Shared pointer to an attitude object for storing IMU data
+   * @note Uses a modified world offset rotation of \f$\pi\f$ radians about the
+   *       roll axis
```
jdcaporale, line 49: add set above

examples/attitude_example.cpp
```diff
@@ -0,0 +1,150 @@
...
+   * @note This is is a shared pointer to a continually updated `Attitude` 
+   *       object.  There is no need to update in the contorl loop itself.
+   */
```
jdcaporale, line 52: control*

examples/attitude_example.cpp
```diff
@@ -0,0 +1,150 @@
...
+    // Get Updated Attitude Representations
+    Eigen::Quaternionf quat = att_->get_att_quat();
+    kodlab::rotations::EulerAngles euler = att_->get_att_euler();
+    Eigen::Matrix3f rot_mat = att_->get_att_rot_mat();
+
+    // Print Attitude Representations to Terminal
+    std::cout << "att_:" << std::endl;
+    att_->PrintAttitude();
+    std::cout << "att_read_only_:" << std::endl;
+    att_read_only_.PrintAttitude();
```
jdcaporale, line 82: These currently should give different answers based on the world offset bug, if not then maybe there is something more insidious

examples/attitude_example.cpp
```diff
@@ -0,0 +1,150 @@
...
+{
+  // Setup Joints
+  std::vector<kodlab::mjbots::JointMoteus> joints;
+  joints.emplace_back(100, 4, 1, -1.3635165, 1, 1);
+  joints.emplace_back(101, 4, -1, 2.688, 5.0 / 3.0, 1);
+  joints.emplace_back(108, 4, 1, -0.4674585, 1, 1);
+
+  // Define Robot Options
+  kodlab::mjbots::ControlLoopOptions options;
+  options.log_channel_name = "attitude_data";
```
jdcaporale, line 129: Maybe world_offset should optionally be set here

include/kodlab_mjbots_sdk/attitude.h
```diff
@@ -0,0 +1,334 @@
...
+ */
+template<typename Scalar = float>
+class Attitude
+{
+
+private:
+  /**
+   * @brief Inputted attitude quaternion before world offset rotation
+   */
+  Eigen::Quaternion<Scalar> att_quat_raw_ = {1, 0, 0, 0};
```
jdcaporale, line 35: IMO "att_..." is redundant, not a huge deal either way and perhaps slightly more correct when compared to the other things in this class, but I think it is ok for quat_,euler_,rot_mat. to be understood to be the attitude, where as

include/kodlab_mjbots_sdk/attitude.h
```diff
@@ -0,0 +1,334 @@
...
+  /**
+   * @brief Attitude quaternion in world frame
+   */
+  Eigen::Quaternion<Scalar> att_quat_ = {1, 0, 0, 0};
+
+  /**
+   * @brief Attitude euler angles in world frame
+   * @note Always follow the default <b>Extrinsic X-Y-Z> convention
+   */
+  mutable kodlab::ValidatedCache<rotations::EulerAngles<Scalar>> att_euler_;
```
jdcaporale, line 46: maybe default to zeros and invalid

include/kodlab_mjbots_sdk/attitude.h
```diff
@@ -0,0 +1,334 @@
...
+
+  /**
+   * @brief Attitude rotation matrix in world frame
+   */
+  mutable kodlab::ValidatedCache<Eigen::Matrix3<Scalar>> att_rot_mat_;
```
jdcaporale, line 51: likewise to identity and invalid

examples/attitude_example.cpp
```diff
@@ -0,0 +1,150 @@
...
+                  const kodlab::mjbots::ControlLoopOptions &options)
+      : MjbotsControlLoop::MjbotsControlLoop(joints, options),
+        att_(robot_->GetAttitudeSharedPtr()),
+        att_read_only_(robot_->GetAttitude())
+  {
+    // Modify world offset rotation to \f$\pi\f$ radians about the roll axis
+    Eigen::Quaternionf world_offset = Eigen::Quaternionf(
+      Eigen::AngleAxisf(M_PI, Eigen::Vector3f::UnitX()));
+    att_->set_world_offset(world_offset);
+    att_read_only_.set_world_offset(world_offset);
```
ethanmusser, line 42: TODO: Clarify that set world offset does nothing for `att_read_only_`

include/kodlab_mjbots_sdk/common_header.h
```diff
@@ -31,4 +33,84 @@ static void EnableCtrlC() {
...
+public:
+  /**
+   * @brief Invalid data constructor
+   */
+  ValidatedCache() : valid_(false) {}
+
+  /**
+   * @brief Valid data constructor
+   */
+  explicit ValidatedCache(const T &data) : data_(data), valid_(true) {}
```
jdcaporale, line 65: Add one more constructor (per comments in attitude class) that allows the setting of a default "zero" or initial value that is valid or not, i.e. explicit ValidatedCache(const T &data, const bool valid) : data_(data), valid_(valid) {}

include/kodlab_mjbots_sdk/attitude.h
```diff
@@ -0,0 +1,334 @@
...
+   * @param accel_in linear acceleration along <code>[x, y, z]</code> axes
+   * @param world_offset_in world frame rotation offset (default identity)
+   */
+  Attitude(const Eigen::Quaternion<Scalar> &att_quat_in,
+           const Eigen::Vector3<Scalar> &ang_rate_in,
+           const Eigen::Vector3<Scalar> &accel_in,
+           const Eigen::Quaternion<Scalar> &world_offset_in = {1, 0, 0, 0})
+      : att_quat_raw_(att_quat_in),
+        att_quat_(world_offset_in * att_quat_in),
+        ang_rate_(ang_rate_in),
```
jdcaporale, line 115: If we have this constructor and the update function to match, we could have "invalid" data in the uncertainty and bias members. Likely we should make each an ValidatedCache or make a "estimator" or "error" struct and make that an invalid cache. Alternatively we could make a simpler class and derive this type of update. I think the intention was to allow the user to use simpler versions in the case we were using a different IMU?

include/kodlab_mjbots_sdk/attitude.h
```diff
@@ -0,0 +1,334 @@
...
+        world_offset_(world_offset_in) {}
+
+  /**
+   * @brief Update this \c Attitude object's data
+   *
+   * @param att_quat_in attitude quaternion
+   * @param ang_rate_in angular velocity about <code>[x, y, z]</code> axes
+   * @param accel_in linear acceleration along <code>[x, y, z]</code> axes
+   */
+  void Update(const Eigen::Quaternion<Scalar> &att_quat_in,
```
jdcaporale, line 181: see above comment

include/kodlab_mjbots_sdk/attitude.h
```diff
@@ -0,0 +1,334 @@
...
+  /**
+   * @return attitude quaternion in world frame
+   */
+  Eigen::Quaternion<Scalar> get_att_quat() const { return att_quat_; }
+
+  /**
+   * @return attitude rotation matrix in world frame
+   */
+  Eigen::Matrix3<Scalar> get_att_rot_mat()
+  {
```
jdcaporale, line 280: as mentioned above, could remove att_ but not necessary

include/kodlab_mjbots_sdk/rotations.h
```diff
@@ -0,0 +1,410 @@
...
+template<typename Scalar = float>
+inline Eigen::Matrix3<Scalar> QuaternionToRotationMatrix(
+    const Eigen::Quaternion<Scalar> &unit_quaternion)
+{
+  return QuaternionToRotationMatrix(unit_quaternion, Scalar(2));
+}
+
+/**
+ * @brief Constructs an euler angle representation from a normalized quaternion
+ *        and a rotation matrix
```
jdcaporale, line 339: should mention that this is specific to roll pitch yaw (and which RPY)

include/kodlab_mjbots_sdk/rotations.h
```diff
@@ -0,0 +1,410 @@
...
+   * @param axes_in array of rotation axes identifiers
+   * @param rot_type_in rotation type (Extrinsic or Intrinsic)
+   */
+  EulerAngles(const Scalar &roll_in,
+              const Scalar &pitch_in,
+              const Scalar &yaw_in,
+              const std::array<int, 3> &axes_in,
+              const EulerRotationType &rot_type_in)
+      : roll(roll_in), pitch(pitch_in), yaw(yaw_in), axes(axes_in),
+        rot_type(rot_type_in) {}
```
jdcaporale, line 125: Are we currently supporting any other convention? if not we may want to remove this and the axes and rot type.
If we did support other conventions, we would need to properly define the constructor to take the angles in the order they were in the convention, e.g. XZY should not construct as EulerAngles euler(X,Y,Z,{0,2,1},EulerAngle::EXTRINSIC)

jdcaporale: Just need a function that maps to roll pitch yaw via the "axes_in"

include/kodlab_mjbots_sdk/rotations.h
```diff
@@ -0,0 +1,410 @@
...
+
+/**
+ * @brief Static assertion that \c EulerAngle object follows X-Y-Z Extrinsic
+ *        Euler angle convention.
+ *
+ * @tparam Scalar numeric type
+ * @param euler Euler angle object
+ */
+template<typename Scalar>
+void ASSERT_DEFAULT_EULER_CONVENTION(const EulerAngles<Scalar> &euler)
```
jdcaporale, line 182: This is a warning to users that currently only XYZ extrinsic exists. Should make an issue to extend Euler angle capabilities to other conventions.

include/kodlab_mjbots_sdk/rotations.h
```diff
@@ -0,0 +1,410 @@
...
+   * @param pitch_in pitch angle
+   * @param yaw_in yaw angle
+   */
+  EulerAngles(const Scalar &roll_in,
+              const Scalar &pitch_in,
+              const Scalar &yaw_in)
+      : roll(roll_in), pitch(pitch_in), yaw(yaw_in) {}
+
+  /**
+   * @brief Construct an Euler Angles object.
```
jdcaporale, line 111: Should be more clear that if it is not XYZ this is meant to act as a POD for your euler angles

include/kodlab_mjbots_sdk/rotations.h
```diff
@@ -0,0 +1,410 @@
...
+
+  /**
+   * @brief Pitch angle (default radians)
+   */
+  Scalar pitch = 0;
+
+  /**
+   * @brief Yaw angle (default radians)
+   */
+  Scalar yaw = 0;
```
jdcaporale, line 59: RPY means we only get taitbryan when we act POD for other conventions 
likely should change to vector if we are going to allow that which we dont have to.

include/kodlab_mjbots_sdk/attitude.h
ethanmusser: Might be that the `Attitude` class is poorly named and should be something more akin to `IMUData`.  But, I think `att_...` is just explicitly identifying the variable data, and matches the convention for the rest of the class variables.

include/kodlab_mjbots_sdk/attitude.h
ethanmusser: Already defaults to invalid.  It could default to zeros for the invalid, cached data, or simply contain garbled data until valid data is provided.  Either is fine really.

include/kodlab_mjbots_sdk/attitude.h
ethanmusser: See above.

include/kodlab_mjbots_sdk/attitude.h
ethanmusser: Yes the intention is for the object to support simpler implementations as well.  I think a note in the constructor comment warning of this intention should suffice.

include/kodlab_mjbots_sdk/attitude.h
ethanmusser: ALso see above.

include/kodlab_mjbots_sdk/attitude.h
ethanmusser: I'm all for leaving it.

examples/attitude_example.cpp
ethanmusser: This can be done with a comment here in the example.

examples/attitude_example.cpp
```diff
@@ -0,0 +1,150 @@
...
+   * @note Uses a modified world offset rotation of \f$\pi\f$ radians about the
+   *       roll axis
+   * @note This is is a shared pointer to a continually updated `Attitude` 
+   *       object.  There is no need to update in the contorl loop itself.
+   */
+  std::shared_ptr<kodlab::Attitude<float>> att_;
+
+  /**
+   * @brief Read-only attitude object for storing IMU data
+   * @note Uses a modified world offset rotation of \f$\pi\f$ radians about the
```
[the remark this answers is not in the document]
ethanmusser: Good catch, comment should be fixed accordingly.

examples/attitude_example.cpp
jdcaporale: Done.

examples/attitude_example.cpp
ethanmusser: Not a bug in world offset, `att_read_only_` is copying the `att_` shared pointer, which has the appropriate offset already set.

examples/attitude_example.cpp
ethanmusser: Agreed, this option would be nice to have.

include/kodlab_mjbots_sdk/common_header.h
ethanmusser: Agreed, this should work for our purposes.  However, a more general class could optionally throw an error.

include/kodlab_mjbots_sdk/attitude.h
jdcaporale: A comment is good starting point, might add issue for dealing with it more carefully in the future

src/mjbots_robot_interface.cpp
ethanmusser: Good point, probably worth moving.

include/kodlab_mjbots_sdk/rotations.h
ethanmusser: This makes sense for generality.  However, will be documented that Tait-Bryan Extrinsic X-Y-Z is the only convention with optimized conversion functions.

include/kodlab_mjbots_sdk/rotations.h
ethanmusser: Additional functionality added in 297e88ce81566e21f82ba644acfb587193045eed makes this redundant.

include/kodlab_mjbots_sdk/rotations.h
ethanmusser: Resolved in 297e88ce81566e21f82ba644acfb587193045eed.

include/kodlab_mjbots_sdk/rotations.h
```diff
@@ -0,0 +1,487 @@
...
+    return axes_out;
+  }
+  
+  /**
+   * @brief Get the angles in extrinsic order
+   * 
+   * @return std::array<float,3> 
+   */
+  std::array<float,3> get_angles_extrinsic()
+  {
```
ethanmusser, line 238: This will not work as written, should read `std::array<float,3> angles_out(angles);`

include/kodlab_mjbots_sdk/rotations.h
```diff
@@ -0,0 +1,487 @@
...
+    std::array<float,3> angles_out(angles_out);
+    if(rot_type==INTRINSIC){
+        std::reverse(angles_out.begin(),angles_out.end());
+    }
+    return angles_out;
+  }
```
ethanmusser, line 244: This function is not templated for generic scalars, will break for any class other than `kodlab::Attitude<float>`

examples/attitude_example.cpp
ethanmusser: t

include/kodlab_mjbots_sdk/attitude.h
ethanmusser: Support for alternative Euler conventions provided in bdfce1322a036db5ad1a8e1a939c31b6042e2025

include/kodlab_mjbots_sdk/attitude.h
ethanmusser: Resolved in bdfce1322a036db5ad1a8e1a939c31b6042e2025

include/kodlab_mjbots_sdk/common_header.h
ethanmusser: This is an issue with prior code, should be fixed in a seperate, formatting PR.  Likely the entire repository should be formatted to match the stated Drake convention in such a PR.

include/kodlab_mjbots_sdk/common_header.h
ethanmusser: Resolved in ced2bfa83422a5806f829c57fa072a9ec6c14f7b

include/kodlab_mjbots_sdk/rotations.h
ethanmusser: Resolved in 3246355abd96ab68f22dd74c68d53d1ee51af0fe

include/kodlab_mjbots_sdk/rotations.h
ethanmusser: Capabilities extended in bdfce1322a036db5ad1a8e1a939c31b6042e2025

include/kodlab_mjbots_sdk/rotations.h
ethanmusser: Clarified in bdfce1322a036db5ad1a8e1a939c31b6042e2025

src/mjbots_robot_interface.cpp
ethanmusser: On second thought, several other accessors are in the implementation as well.  This move should be addressed in a seperate PR.

examples/attitude_example.cpp
ethanmusser: Feature implemented in 27d831239563df30c894c879e6625264fb52358a

examples/attitude_example.cpp
ethanmusser: No longer relevant

examples/attitude_example.cpp
ethanmusser: Resolved in 89051060028b8a55d24cfc934a3f07b91d7f66b2

examples/attitude_example.cpp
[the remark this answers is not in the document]
ethanmusser: Resolved in 89051060028b8a55d24cfc934a3f07b91d7f66b2.  Additionally, can be set via `options` struct since 27d831239563df30c894c879e6625264fb52358a.

include/kodlab_mjbots_sdk/attitude.h
ethanmusser: Resolved in ced2bfa83422a5806f829c57fa072a9ec6c14f7b

include/kodlab_mjbots_sdk/attitude.h
ethanmusser: Resolved in ced2bfa83422a5806f829c57fa072a9ec6c14f7b

include/kodlab_mjbots_sdk/rotations.h
ethanmusser: Resolved in 297e88ce81566e21f82ba644acfb587193045eed

ethanmusser:
Per the discussions above, I think this is ready to be merged in.

ethanmusser:
Pending the remaining, unresolved discussions above (which are minor in scope), I think this is ready to be merged in.

ShaneRozenLevy:
You should add some comments to readme on how to use it.

src/mjbots_robot_interface.cpp
```diff
@@ -80,6 +82,14 @@ MjbotsRobotInterface::MjbotsRobotInterface(const std::vector<JointMoteus> &joint
   moteus_options.imu_mounting_deg = imu_mounting_deg;
   moteus_interface_ = std::make_shared<::mjbots::moteus::Pi3HatMoteusInterface>(moteus_options);
 
+  // Initialize attitude shared pointer
+  attitude_ = std::make_shared<::kodlab::Attitude<float>>();
+  kodlab::rotations::EulerAngles<float> imu_world_offset =
```
ShaneRozenLevy, line 87: we probably want to make a free function that does this conversion. We can think about where to include it.

include/kodlab_mjbots_sdk/rotations.h
```diff
@@ -0,0 +1,590 @@
...
+const std::array<int, 3> DEFAULT_EULER_ANGLE_AXES = {0, 1, 2};
+const EulerRotationType DEFAULT_EULER_ANGLE_ROTATION_TYPE = EXTRINSIC;
+
+/**
+ * @brief Euler angle rotation object.
+ * @note Default convention is <b>X-Y-Z Extrinsic</b> Euler angles.
+ *
+ * @tparam Scalar[optional] numeric type
+ */
+template<typename Scalar = float>
```
ShaneRozenLevy, line 41: My concern with this block of code is that if we are not using this right now and we have not tested it, it could easily be bugged.

ShaneRozenLevy (commented):
Y'all should read this: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines

include/kodlab_mjbots_sdk/attitude.h
ethanmusser: Class name changed to `IMUData` and header name changed to `imu_data.h` in 2fe609067d9ac2f834dc243b9f5e117bbf844c36. Class members and accessors changed to remove `att_...` in 88098f1aa2c90661cd905c056e1c3bd4b1044d2b.  Marking this as resolved.

include/kodlab_mjbots_sdk/attitude.h
ethanmusser: Class members and accessors changed to remove `att_...` in 88098f1aa2c90661cd905c056e1c3bd4b1044d2b.  Marking this as resolved.

examples/attitude_example.cpp
```diff
@@ -0,0 +1,133 @@
...
+public:
+  using MjbotsControlLoop::MjbotsControlLoop;
+
+private:
+  /**
+   * @brief Shared pointer to an attitude object for storing IMU data
+   * @note This is is a shared pointer to a continually updated `Attitude`
+   *       object.  There is no need to update in the control loop itself.
+   */
+  std::shared_ptr<kodlab::Attitude<float>> att_ = robot_->GetAttitudeSharedPtr();
```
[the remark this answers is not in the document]
ethanmusser: Marked `const` in 11fe11025131aa37d376ffd59930c831a27d7e56.  As you said, this is an example, so still not using `auto`.  Marking as resolved.

include/kodlab_mjbots_sdk/mjbots_robot_interface.h
```diff
@@ -146,7 +149,13 @@ class MjbotsRobotInterface {
    * @brief accessor for the attitude of the robot
    * @return the attitude object for the robot
```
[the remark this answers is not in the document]
ethanmusser: Fixed in 84281c314bee55c3a70a33717634439ad90780c8.  Marking as resolved.

include/kodlab_mjbots_sdk/mjbots_robot_interface.h
```diff
@@ -146,7 +149,13 @@ class MjbotsRobotInterface {
...
    */
-  ::mjbots::pi3hat::Attitude GetAttitude();
+  const ::kodlab::Attitude<float>& GetAttitude();
+
+  /*!
+   * @brief accessor for the attitude of the robot
+   * @return the attitude shared pointer for the robot
+   */
+  std::shared_ptr<::kodlab::Attitude<float>> GetAttitudeSharedPtr();
```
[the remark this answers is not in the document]
ethanmusser: Yes, we do.  Remedied in 84281c314bee55c3a70a33717634439ad90780c8.  Marking as resolved.

examples/attitude_example.cpp
```diff
@@ -0,0 +1,133 @@
...
+  /**
+   * @brief Gathers attitude data and sets command torques to zero
+   */
+  void CalcTorques() override
+  {
+    // Update Read-Only Attitude
+    att_read_only_ = robot_->GetAttitude();
+
+    // Read Various Attitude Information from `att_`
+    Eigen::Quaternionf quat = att_->get_att_quat();    // attitude quaternion
```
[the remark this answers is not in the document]
ethanmusser: For the example, we are trying to be very explicit about what types the functions return.  For general use, I completely agree.  Marked as `const` in 84281c314bee55c3a70a33717634439ad90780c8.  Marking as resolved.

include/kodlab_mjbots_sdk/attitude.h
```diff
@@ -0,0 +1,363 @@
...
+#include "kodlab_mjbots_sdk/common_header.h"
+#include "kodlab_mjbots_sdk/rotations.h"
+#include <iostream>
+#include <iomanip>
+
+namespace kodlab
+{
+
+/**
+ * @brief Attitude object storing inertial data.
```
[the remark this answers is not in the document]
ethanmusser: I think this is sufficiently clear now that the class is named `IMUData` (change made in 2fe609067d9ac2f834dc243b9f5e117bbf844c36).

include/kodlab_mjbots_sdk/attitude.h
```diff
@@ -0,0 +1,363 @@
...
+  /**
+   * @brief Attitude quaternion in world frame
+   */
+  Eigen::Quaternion<Scalar> att_quat_ = {1, 0, 0, 0};
+
+  /**
+   * @brief Attitude euler angles in world frame
+   * @note Always follow the default <b>Extrinsic X-Y-Z> convention
+   */
+  mutable kodlab::ValidatedCache<rotations::EulerAngles<Scalar>>
```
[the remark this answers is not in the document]
ethanmusser: The `ValidatedCache` object internally modifies a `bool` validity flag and overwrites cached data if called.  In `const` usages, we still want these values to be computed and stored using the `const` attitude quaternion upon which all of these are computed. 
 Marking mutable allows the internal data and flag to be modified appropriately.

include/kodlab_mjbots_sdk/attitude.h
```diff
@@ -0,0 +1,363 @@
...
+   * @param ang_rate_bias_in gyroscope bias about <code>[x, y, z]</code> axes
+   * @param att_uncertainty_in attitude uncertainty quaternion
+   * @param ang_bias_uncertainty_in gyroscope bias uncertainty about
+   *                                <code>[x, y, z]</code> axes
+   * @param world_offset_in world frame rotation offset (default identity)
+   */
+  Attitude(const Eigen::Quaternion<Scalar> &att_quat_in,
+           const Eigen::Vector3<Scalar> &ang_rate_in,
+           const Eigen::Vector3<Scalar> &accel_in,
+           const Eigen::Vector3<Scalar> &ang_rate_bias_in,
```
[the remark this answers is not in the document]
ethanmusser: This is so that the `IMUData` class world offset can be easily modified both in cases where only {attitude, ang. velocity, acceleration} are used, as well as those usages including uncertainty and biases.

include/kodlab_mjbots_sdk/common_header.h
```diff
@@ -31,4 +33,89 @@ static void EnableCtrlC() {
   sigaction(SIGINT, &sigIntHandler, NULL);
   CTRL_C_DETECTED = false;
 }
+
+/**
+ * @brief Simple object wrapper that caches the latest value until invalidated
+ * @tparam T typename
+ */
+template<typename T>
```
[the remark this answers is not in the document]
ethanmusser: `ValidatedCache` provides convenience functions for setting and validating, and invalidating data that `std::optional` does not have.  It also permits accessing of invalid (in our case, out-of-date) data and overloads the assignment operator to permit implicit conversion to the stored type. Also, `std::optional` requires handling of the `std::bad_optional_access` exception.

That said, the internals of `ValidatedCache` could be replaced with `std::optional`.

src/mjbots_robot_interface.cpp
ethanmusser: Agreed, we need a header of Mjbots free functions somewhere in the `mjbots` namespace.  However, I think this may deserve a separate PR once this one goes through.

include/kodlab_mjbots_sdk/attitude.h
```diff
@@ -0,0 +1,363 @@
...
+
+  /**
+   * @brief Construct a new \c Attitude object with populated data.
+   *
+   * @param att_quat_in attitude quaternion
+   * @param ang_rate_in angular velocity about <code>[x, y, z]</code> axes
+   * @param accel_in linear acceleration along <code>[x, y, z]</code> axes
+   * @param world_offset_in world frame rotation offset (default identity)
+   */
+  Attitude(const Eigen::Quaternion<Scalar> &att_quat_in,
```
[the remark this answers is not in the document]
ethanmusser: These are references to incoming IMU data (in the current implementation, data from the pi3hat registers.

include/kodlab_mjbots_sdk/attitude.h
```diff
@@ -0,0 +1,363 @@
...
+{
+
+private:
+  /**
+   * @brief Inputted attitude quaternion before world offset rotation
+   */
+  Eigen::Quaternion<Scalar> att_quat_raw_ = {1, 0, 0, 0};
+
+  /**
+   * @brief Attitude quaternion in world frame
```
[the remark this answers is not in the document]
ethanmusser: Clarified in 49433214581affbd65dd7e79cf233948c83bfd16.  Marking as resolved.

include/kodlab_mjbots_sdk/attitude.h
```diff
@@ -0,0 +1,363 @@
...
+  /**
+   * @brief World frame rotation offset
+   * @note Useful to change the direction of gravity
+   */
+  Eigen::Quaternion<Scalar> world_offset_ = {1, 0, 0, 0};
+
+  /**
+   * @brief Invalidates cached data
+   */
+  void InvalidateCached()
```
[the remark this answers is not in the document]
ethanmusser: Renamed in 49433214581affbd65dd7e79cf233948c83bfd16.  Marking as resolved.

include/kodlab_mjbots_sdk/attitude.h
```diff
@@ -0,0 +1,363 @@
...
+            pi_att.attitude_uncertainty.y, pi_att.attitude_uncertainty.z},
+           {pi_att.bias_uncertainty_dps.x, pi_att.bias_uncertainty_dps.y,
+            pi_att.bias_uncertainty_dps.z});
+  }
+
+  /**
+   * @brief Prints attitude, angular velocity, and linear acceleration in
+   * various representations
+   */
+  void PrintAttitude()
```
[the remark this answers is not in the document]
ethanmusser: Added in 49433214581affbd65dd7e79cf233948c83bfd16.  Marking as resolved.

include/kodlab_mjbots_sdk/rotations.h
ethanmusser: All untested portions have been commented as of 1c333dee37947609c1b04aec046276f38b3347ed.  Marking as resolved.

include/kodlab_mjbots_sdk/attitude.h
jdcaporale: Perhaps an attitude specific struct could exist in the future for added clarity, but for now this LGTM.

include/kodlab_mjbots_sdk/attitude.h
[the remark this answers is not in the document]
jdcaporale: Eigen recommends passing Eigen Types this way to preserve their alignment and then to use the copy constructor (used in the initalizers).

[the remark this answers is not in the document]
ethanmusser: > Are these initial values, or are the references to a different data source?

I misunderstood the question in my original response.  @jdcaporale is correct.  See [this link](https://eigen.tuxfamily.org/dox/group__TopicPassingByValue.html) for details.

include/kodlab_mjbots_sdk/attitude.h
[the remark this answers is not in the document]
jdcaporale: This is another case pushing for having a simple attitude struct inside or along side of imu_data so as to separate those parts of the class intuitively. [Prev](https://github.com/KodlabPenn/kodlab_mjbots_sdk/pull/16/files/61bd4eed34b1371f16943b3f609522a114a45c50#r935094001)

include/kodlab_mjbots_sdk/common_header.h
jdcaporale: Sure.

include/kodlab_mjbots_sdk/common_header.h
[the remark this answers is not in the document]
jdcaporale: I think that unless we ask for a default value, std::optional would force us to handle its NULL capability outside this class or to error out which could be dangerous. So for now I don't think it is what we want. It is better used for when data is actually optional and not for this caching in particular.

include/kodlab_mjbots_sdk/attitude.h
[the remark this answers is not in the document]
ethanmusser: @jdcaporale make an issue out of this.

src/mjbots_robot_interface.cpp
jdcaporale: Agreed, let's an issue to that end.

jdcaporale: #29

include/kodlab_mjbots_sdk/imu_data.h
```diff
@@ -22,6 +22,16 @@ namespace kodlab
 
 /**
  * @brief IMUData object storing inertial data.
+ * @note The body, world, and modified world frames are referred to throughout
+ *       this class' documentation. The body frame is fixed to the robot body
+ *       and offset from the IMU/Pi3Hat frame based on the Mjbots control loop
+ *       options `imu_mounting_deg` parameters. The world frame describes a
+ *       fixed inertial frame which is defined at robot startup. This frame
+ *       always has \f$\hat{z}\f$ pointing in the direction of gravity. The
```
jdcaporale, line 30: This should not mention MJBots if we are trying to keep things modular. #29 is a useful place to start

include/kodlab_mjbots_sdk/imu_data.h
```diff
@@ -0,0 +1,373 @@
...
+        world_offset_(world_offset_in) {}
+
+  /**
+   * @brief Construct a new \c IMUData object from a
+   *        \c mjbots::pi3hat::IMUData object
+   *
+   * @param pi_att attitude object
+   * @param world_offset_in world frame rotation offset (default identity)
+   */
+  IMUData(const ::mjbots::pi3hat::Attitude &pi_att,
```
jdcaporale, line 167: This should be fixed in #29 along with [above](https://github.com/KodlabPenn/kodlab_mjbots_sdk/pull/16/files#r935106685)

include/kodlab_mjbots_sdk/rotations.h
```diff
@@ -0,0 +1,603 @@
...
+ *       and verification for robotics" (2019)
+ * @link https://github.com/RobotLocomotion/drake/blob/5022e3832e90f988f1b6d3015b2794ea26e527b1/math/roll_pitch_yaw.cc#L112 @endlink
+ *
+ * @tparam Scalar[optional] numeric type (default float)
+ * @param quaternion
+ * @param R
+ * @return EulerAngles<Scalar>
+ */
+template<typename Scalar = float>
+EulerAngles<Scalar> QuaternionAndRotationMatrixToEulerAngles(
```
jdcaporale, line 544: These last two functions should be renamed to match above. Namely
```suggestion
EulerAngles<Scalar> QuaternionAndRotationMatrixToDefaultEulerAngles(
```

include/kodlab_mjbots_sdk/rotations.h
```diff
@@ -0,0 +1,603 @@
...
+
+/**
+ * @brief Constructs euler angles from a normalized quaternion
+ *
+ * @tparam Scalar[optional] numeric type (default float)
+ * @param unit_quaternion
+ * @return EulerAngles<Scalar>
+ */
+template<typename Scalar = float>
+EulerAngles<Scalar> QuaternionToEulerAngles(
```
jdcaporale, line 595:
```suggestion
EulerAngles<Scalar> QuaternionToDefaultEulerAngles(
```

ethanmusser: Resolved in d3d867197e2aa610a80dc502b3811b3b1d9302de

include/kodlab_mjbots_sdk/rotations.h
ethanmusser: Resolved in d3d867197e2aa610a80dc502b3811b3b1d9302de

jdcaporale (approved):
LGTM

PR #18, opened 2022-07-13
Moves JointBase zero offset from servo-space to joint-space

ethanmusser:
Zero offset was mistakenly computed in servo-space.  PR defines zero offset as being in joint space.


jdcaporale (approved):
This is a better way to do it. 
Was this way in dev, good PR.
LGTM

PR #20, opened 2022-07-14
Macro loggers

ethanmusser:
Provides macro logging infrastructure in `log.h`.  Logging macros document relevant status information with `fprintf`-like syntax.  Simpler to use than the present `ostream` method.

### Possible Additions

- Blurb detailing usage in the `README.md` or wiki
- Logging to a file
- Adding time to displayed output
- Implementing timing capabilities

### Usage

The input 
```cpp
LOG_DEBUG("Debug message.");
LOG_INFO("%s message.", "Info");
LOG_WARN("%s %s.", "Warn", "message");
LOG_ERROR("Error message. Error code: %d.", 10);
LOG_FATAL("Fatal message.");

LOG_IF_DEBUG(true, "This debug message will print.");
LOG_IF_INFO(false, "This info message won't print");
LOG_IF_WARN(true, "String substitution works in this %s if statement.", "warn");
LOG_IF_ERROR(0 < 1, "Error message will print.  (0 < 1) = %d", 0 < 1);
LOG_IF_FATAL(1 == 2, "Fatal message will not print.");
```
produces the following output when called in the `/home/ethan/kodlab_mjbots_sdk/examples/spin_joints_example.cpp` `main` function on lines spanning 61 - 70.
Screenshot from 2022-07-27 18-08-23

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,51 @@ To configure the soft Start, set the `options.max_torque` and `options.soft_star
 max torque is the maximum torque per motor and the soft Start duration is how long the torque ramp should last
 in iterations of the control loop. 
 
+## Console Logging
+The `log.h` header provides a set of debug logging macros with adjustable
+logging severity levels.  In order of increasing severity, the levels are 
+`DEBUG`, `INFO`, `WARN`, `ERROR`, and `FATAL`.  
+
+The minimum level for console output is set by defining `LOG_MIN_SEVERITY` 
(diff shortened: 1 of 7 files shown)


ShaneRozenLevy (changes requested):
LGTM a few minor blockers:

1. Add a few paragraphs to the readme with how to use this
2. Add this to an example
3. Add a screenshot to the pr with what it looks like

In the future you should add a stack trace.

ShaneRozenLevy (approved):
LGTM

PR #21, opened 2022-07-17
DRAFT Simplifying future work with architecture change

jdcaporale:
Most notably separated a RobotInterface class and the MjbotsRobotInterface class (whose name should likely change, perhaps to MjbotsInterface) in the control loop. This allowed the control loop to more easily take a RobotInterfaceDerived class as a template argument. Need to make some decisions on the names of function members for the control loop. Changes to existing examples and code should be minor, but this will certainly move some of the member names around. 

NOTE: If you try to build this, you will note that spin_joints_example is broken for now as an example of the types of errors/changes this makes and all the other of the examples are commented out for this DRAFT.


jdcaporale:
Closing to build cleaner more accurate PR

PR #22, opened 2022-07-17
Mjbots interface composition for pr

jdcaporale:
Most notably separated a RobotInterface class and the MjbotsRobotInterface class (whose name should likely change, perhaps to MjbotsInterface) in the control loop. This allowed the control loop to more easily take a RobotInterfaceDerived class as a template argument. Need to make some decisions on the names of function members for the control loop. Changes to existing examples and code should be minor, but this will certainly move some of the member names around.

### DIFF
--- a/examples/robot_example.cpp
+++ b/examples/robot_example.cpp
@@ -58,50 +58,16 @@ class SimpleRobotControlLoop : public kodlab::mjbots::MjbotsControlLoop<ManyMoto
     }
 };
 
-/*!
- * @brief Helper class to wrap the vector of shared pointers of JointBaseDerived
- * 
- * @tparam joint_type must be JointBaseDerived type
- */
-template <class joint_type>
-class JointSharedVector
-{
-    static_assert(std::is_base_of<JointBase, joint_type>::value); // check that the joint_type is derived from JointBase
-    std::vector<std::shared_ptr<joint_type>> v_;  // internal vector of shared pointers 
-
-public:
-    /*!
-     * @brief constructs a shared_ptr joint_type and appends it to the internal vector
-     * 
-     * @tparam Args parameter pack for passing through the joint constructor
-     * @param args a set of args that match the joint_type constructor
-     */
-    template <typename... Args>
-    void addJoint(Args... args)
-    {
-        v_.push_back(std::make_shared<joint_type>(args...));
-    }
-
-    /*!
-     * @brief implicit conversion to vector that returns a copy of the shared_ptr's
-     * 
-     * @return std::vector<std::shared_ptr<joint_type>> 
-     */
-    operator std::vector<std::shared_ptr<joint_type>>() { return v_; }
-};
-
 int main(int argc, char **argv)
 {
 
-    // Setup joints with a std::vector
-    // std::vector<kodlab::mjbots::JointMoteus> joints;
-    // joints.emplace_back(100, 4, 1, 0,   1, 0);
-    // joints.emplace_back(101, 4,-1, 0, 5.0/3.0, 0);
-
-    // Setup joints with JointSharedVector
-    JointSharedVector<kodlab::mjbots::JointMoteus> joints;
-    joints.addJoint(100, 4, 1, 0, 1, 0);
-    joints.addJoint(101, 4, -1, 0, 5.0 / 3.0, 0);
+    // Setup joints with a std::vector or JointSharedVector class
+    std::vector<kodlab::mjbots::JointMoteus> joints;
+    joints.emplace_back(100, 4, 1, 0,   1, 0);
+    joints.emplace_back(101, 4,-1, 0, 5.0/3.0, 0);
+    // JointSharedVector<kodlab::mjbots::JointMoteus> joints;
+    // joints.addJoint(100, 4, 1, 0, 1, 0);
+    // joints.addJoint(101, 4, -1, 0, 5.0 / 3.0, 0);
 
     // Define robot options
     kodlab::mjbots::ControlLoopOptions options;
--- a/examples/robot_example.cpp
+++ b/examples/robot_example.cpp
@@ -15,7 +15,7 @@
 #include "kodlab_mjbots_sdk/common_header.h"
 #include "ManyMotorLog.hpp"
 #include "ModeInput.hpp"
-#include "kodlab_mjbots_sdk/robot_interface.h"
+#include "kodlab_mjbots_sdk/robot_base.h"
 #include "kodlab_mjbots_sdk/mjbots_control_loop.h"
 
 #include "examples/simple_robot.h"
--- a/examples/robot_example.cpp
+++ b/examples/robot_example.cpp
@@ -1,0 +1,98 @@
+
+// #include "kodlab_mjbots_sdk/robot_control_loop.h"
+// #include "kodlab_mjbots_sdk/behavior.h"
+// #include "kodlab_mjbots_sdk/robot_super.h"
+#include <vector>
+#include <iostream>
+#include <memory>
+#include "kodlab_mjbots_sdk/common_header.h"
+#include "ManyMotorLog.hpp"
+#include "ModeInput.hpp"
+#include "kodlab_mjbots_sdk/robot_interface.h"
+#include "kodlab_mjbots_sdk/mjbots_control_loop.h"  
+
+#include "examples/simple_robot.h"
+
+class MjbotsSimpleRobot : public SimpleRobot, public kodlab::mjbots::MjbotsRobotInterface
+{   
+public:
+    int mode = 0;
+    MjbotsSimpleRobot( std::vector<std::shared_ptr<kodlab::mjbots::JointMoteus>> joints_in , 
+                        kodlab::mjbots::ControlLoopOptions &options)
+    : RobotInterface(joints_in,options.max_torque,options.soft_start_duration),
+      kodlab::mjbots::MjbotsRobotInterface(joints_in, options.realtime_params,options.soft_start_duration,options.max_torque,options.imu_mounting_deg,options.attitude_rate_hz)
+      {}
+
+};
+
+class SimpleRobotControlLoop : public kodlab::mjbots::MjbotsControlLoop<ManyMotorLog,ModeInput> {
+  using MjbotsControlLoop::MjbotsControlLoop;
+
+  void CalcTorques() override {
+    std::vector<float> torques(num_motors_, 0);
+    robot_->SetTorques(torques);
+  }  
+  void PrepareLog() override {
+      for (int servo = 0; servo < num_motors_; servo++) {
+          log_data_.positions[servo] = robot_->GetJointPositions()[servo];
+          log_data_.velocities[servo] = robot_->GetJointVelocities()[servo];
+          log_data_.modes[servo] = static_cast<int>(robot_->GetJointModes()[servo]);
+          log_data_.torques[servo] = robot_->GetJointTorqueCmd()[servo];
+      }
+      for (int servo = num_motors_; servo < 13; servo++) {
+          log_data_.positions[servo] = 0;
+          log_data_.velocities[servo] = 0;
+          log_data_.modes[servo] = 0;
+          log_data_.torques[servo] = 0;
+      }
+  } 
+  
+  void ProcessInput() override {
+    int behavior = 0;
+      behavior = lcm_sub_.data_.mode;
+      std::cout << "Switching to behavior " << behavior  << std::endl;
+      // If the kill robot mode is detected kill robot using CTRL_C flag handler.
+      if (behavior  == -1){ //KILL_ROBOT
+          kodlab::CTRL_C_DETECTED = true;
+      }
+  }
+};
+template<class joint_type>
+class JointSharedVector {
+    std::vector<std::shared_ptr<joint_type>> v_;
+  public:
+    template<typename... Args>
+    void addJoint(Args... args){
+      v_.push_back(std::make_shared<joint_type>(args...));
+    }
+    operator std::vector<std::shared_ptr<joint_type>> () { return v_; }
+};
+
+
+
+int main(int argc, char **argv) {
+
+
+  //Setup joints
+  // std::vector<kodlab::mjbots::JointMoteus> joints2;
+  // joints2.emplace_back(100, 4, 1, 0,   1, 0);
+  // joints2.emplace_back(101, 4,-1, 0, 5.0/3.0, 0);
+
+  JointSharedVector<kodlab::mjbots::JointMoteus> joints;
+  joints.addJoint(100, 4, 1, 0,   1, 0);
+  joints.addJoint(101, 4,-1, 0, 5.0/3.0, 0);
+
+  // Define robot options
+  kodlab::mjbots::ControlLoopOptions options;
+  options.log_channel_name = "motor_data";
+  options.frequency = 1000;
+  options.realtime_params.main_cpu = 3;
+  options.realtime_params.can_cpu  = 2;
+  MjbotsSimpleRobot rob(joints, options);
+
+  // Create control loop
+  // Starts the loop, and then join it
+  SimpleRobotControlLoop simple_robot(std::make_shared<MjbotsSimpleRobot>(std::move(rob)),options);
+  simple_robot.Start();
+  simple_robot.Join();
+}
--- a/examples/robot_example.cpp
+++ b/examples/robot_example.cpp
@@ -13,41 +13,7 @@
 
 #include "examples/simple_robot.h"
 
-class MjbotsSimpleRobot : public kodlab::mjbots::MjbotsRobotInterface, public SimpleRobot
-{
-public:
-    MjbotsSimpleRobot(std::vector<std::shared_ptr<kodlab::mjbots::JointMoteus>> joints_in,
-                      kodlab::mjbots::ControlLoopOptions &options)
-        : RobotInterface(joints_in, options.max_torque, options.soft_start_duration),
-          kodlab::mjbots::MjbotsRobotInterface(joints_in,
-                                               options.realtime_params,
-                                               options.soft_start_duration,
-                                               options.max_torque,
-                                               options.imu_mounting_deg,
-                                               options.attitude_rate_hz)
-    {
-        mode = 0;
-    }
-
-    MjbotsSimpleRobot(std::vector<std::shared_ptr<kodlab::mjbots::JointMoteus>> joints_in,
-                      const kodlab::mjbots::RealtimeParams &realtime_params,
-                      int soft_start_duration,
-                      float robot_max_torque,
-                      ::mjbots::pi3hat::Euler imu_mounting_deg,
-                      int imu_rate_hz)
-        : RobotInterface(joints_in, robot_max_torque, soft_start_duration),
-          kodlab::mjbots::MjbotsRobotInterface(joints_in,
-                                               realtime_params,
-                                               soft_start_duration,
-                                               robot_max_torque,
-                                               imu_mounting_deg,
-                                               imu_rate_hz)
-    {
-        mode = 0;
-    }
-};
-
-class SimpleRobotControlLoop : public kodlab::mjbots::MjbotsControlLoop<ManyMotorLog, ModeInput, MjbotsSimpleRobot>
+class SimpleRobotControlLoop : public kodlab::mjbots::MjbotsControlLoop<ManyMotorLog, ModeInput, SimpleRobot>
 {
     using MjbotsControlLoop::MjbotsControlLoop;
 
@@ -61,7 +27,7 @@ class SimpleRobotControlLoop : public kodlab::mjbots::MjbotsControlLoop<ManyMoto
         {
             log_data_.positions[servo] = robot_->GetJointPositions()[servo];
             log_data_.velocities[servo] = robot_->GetJointVelocities()[servo];
-            log_data_.modes[servo] = static_cast<int>(robot_->GetJointModes()[servo]);
+            log_data_.modes[servo] = static_cast<int>(mjbots_interface_->GetJointModes()[servo]);
             log_data_.torques[servo] = robot_->GetJointTorqueCmd()[servo];
         }
         for (int servo = num_motors_; servo < 13; servo++)
--- a/examples/robot_example.cpp
+++ b/examples/robot_example.cpp
@@ -24,7 +24,7 @@ class SimpleRobotControlLoop : public kodlab::mjbots::MjbotsControlLoop<ManyMoto
 {
     using MjbotsControlLoop::MjbotsControlLoop;
 
-    void CalcTorques() override
+    void Update() override
     {
         robot_->Update();
     }
--- a/examples/robot_example.cpp
+++ b/examples/robot_example.cpp
@@ -1,7 +1,14 @@
-
-// #include "kodlab_mjbots_sdk/robot_control_loop.h"
-// #include "kodlab_mjbots_sdk/behavior.h"
-// #include "kodlab_mjbots_sdk/robot_super.h"
+/*!
+ * @file robot_example.cpp
+ * @author J. Diego Caporale <jdcap@seas.upenn.edu>
+ * @brief Example that uses SimpleRobot(a RobotInterfaceDerived class defined in a seperate header file) separating the 
+ *        controller, state update, and state machine from the lower lever control loop. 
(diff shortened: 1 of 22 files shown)


examples/robot_example.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+#include "examples/simple_robot.h"
+
+class SimpleRobotControlLoop : public kodlab::mjbots::MjbotsControlLoop<ManyMotorLog, ModeInput, SimpleRobot>
+{
+    using MjbotsControlLoop::MjbotsControlLoop;
+
+    void CalcTorques() override
+    {
+        robot_->Update();
```
jdcaporale, line 22: Not sure if this is the interaction we want or if this should be hidden in robot unless you want an override

examples/robot_example.cpp
```diff
@@ -0,0 +1,94 @@
...
+        std::cout << "Switching to behavior " << robot_->mode << std::endl;
+        // If the kill robot mode is detected kill robot using CTRL_C flag handler.
+        if (robot_->mode == robot_->KILL_ROBOT)
+        { // KILL_ROBOT
+            kodlab::CTRL_C_DETECTED = true;
+        }
+    }
+};
+
+template <class joint_type>
```
jdcaporale, line 54: This class template could be moved to the joint base class probably though templating would get weird. Useful way to used shared pointers without forcing c++syntax. Probably a few other functions that could be moved in here like the getters and stuff, but right now this is implicitly converted to a vector of shared pointers.

jdcaporale: I thought about deriving std::vector but decided it was a bad idea with lots of ways it could go wrong.

examples/robot_example.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+int main(int argc, char **argv)
+{
+
+    // Setup joints
+    // std::vector<kodlab::mjbots::JointMoteus> joints;
+    // joints.emplace_back(100, 4, 1, 0,   1, 0);
+    // joints.emplace_back(101, 4,-1, 0, 5.0/3.0, 0);
+
+    JointSharedVector<kodlab::mjbots::JointMoteus> joints;
```
jdcaporale, line 77: Some of the simplifications that happened in the later commits mean this new class while useful as mentioned above may no longer be necessary.

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -42,14 +44,24 @@ struct ControlLoopOptions {
...
    */
   MjbotsControlLoop(std::vector<kodlab::mjbots::JointMoteus> joints, const ControlLoopOptions &options);
+  /*!
+   * \overload 
+   */
+  MjbotsControlLoop(std::vector<std::shared_ptr<kodlab::mjbots::JointMoteus>> joints, const ControlLoopOptions &options);
+  /*!
+   * \overload 
+   */
+  MjbotsControlLoop(std::shared_ptr<RobotClass>robot_in, const ControlLoopOptions &options);
```
jdcaporale, line 64: Needs comment for robot input

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -61,7 +73,7 @@ class MjbotsControlLoop : public AbstractRealtimeObject {
   /*!
    * @brief function to be implemented by child. Must set torques in the robot class
    */
-  virtual void CalcTorques() = 0;
+  virtual void CalcTorques() {robot_->Update();}
```
jdcaporale, line 76: this interaction could change (as mentioned above)

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -92,9 +104,15 @@ class MjbotsControlLoop : public AbstractRealtimeObject {
...
 
-  std::shared_ptr<MjbotsRobotInterface>
-      robot_;   /// ptr to the robot object, if unique causes many issues, also should be
-  /// initialized inside thread
+  /*!
+   * @brief Construct a new Setup Options object
+   * 
+   * @param options 
+   */
+  void SetupOptions(const ControlLoopOptions &options);
```
jdcaporale, line 112: useful for writing new constructors

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -109,16 +127,48 @@ class MjbotsControlLoop : public AbstractRealtimeObject {
...
+  SetupOptions(options);
+}
+
+template<class log_type, class input_type, class robot_type>
+MjbotsControlLoop<log_type, input_type, robot_type>::MjbotsControlLoop(std::shared_ptr<robot_type>robot_in, const ControlLoopOptions &options)
+  : AbstractRealtimeObject(options.realtime_params.main_rtp, options.realtime_params.can_cpu),
     lcm_sub_(options.realtime_params.lcm_rtp, options.realtime_params.lcm_cpu, options.input_channel_name) {
+  // Create robot object
+  robot_ = robot_in;
+  mjbots_interface_ = std::make_shared<kodlab::mjbots::MjbotsRobotInterface>( robot_->joints,
```
jdcaporale, line 157: This composition one of the important changes along with the additions of a RobotInterface class

lcm_types/mode_input.lcm
```diff
@@ -0,0 +1,4 @@
+struct ModeInput
+{
+  int16_t    mode;
```
jdcaporale, line 3: Should probably be a different PR?

src/mjbots_robot_interface.cpp
```diff
@@ -47,28 +47,19 @@ ::mjbots::moteus::QueryResult MjbotsRobotInterface::Get(const std::vector<::mjbo
...
   for( auto & j: joints){
-    positions_.push_back( j->get_position_reference() );
-    velocities_.push_back( j->get_velocity_reference() );
-    torque_cmd_.push_back( j->get_servo_torque_reference()   ); 
     modes_.push_back(j->get_mode_reference());
   }
-  num_servos_ = joints.size();
-  for (size_t i = 0; i < num_servos_; ++i)
+  
+  for (size_t i = 0; i < num_joints_; ++i)
```
jdcaporale, line 62: Changed for consistency still num_motors in controlLoop, should be the same everywhere?

lcm_types/mode_input.lcm
ethanmusser: Agreed, should probably rebase a81f6688a616721fc89d664e42eecd35c83525ee off of fd4b851c454f5986c7b01015873a63c63686b3dd

src/robot_interface.cpp
```diff
@@ -0,0 +1,52 @@
+/*!
+ * @file robot_interface.cpp
```
ethanmusser, line 2: This class is mostly accessors and mutators.  Consider consolidating these portions of the implementation into `inline` functions in the class definition in `robot_interface.h`.

src/robot_interface.cpp
```diff
@@ -0,0 +1,52 @@
...
+  std::vector<float>torques(torque_cmd_.begin(), torque_cmd_.end());
+  return torques;
+}
+void kodlab::RobotInterface::SetTorques(std::vector<float> torques) {
+  soft_start_.ConstrainTorques(torques, cycle_count_);
+  float torque_cmd;
+  for (int joint_ind = 0; joint_ind < num_joints_; joint_ind++) {
+    torque_cmd = joints[joint_ind]->UpdateTorque(torques[joint_ind]);
+  }
+}
```
ethanmusser, line 51: This class is mostly accessors and mutators.  Consider consolidating these portions of the implementation into `inline` functions in the class definition in `robot_interface.h`.

include/examples/simple_robot.h
```diff
@@ -0,0 +1,14 @@
+#include "kodlab_mjbots_sdk/robot_interface.h"
```
ethanmusser, line 1: Missing file header for Doxygen

examples/robot_example.cpp
ethanmusser: I think the latter.  This would permit the override if desired (and support backwards compatability) while encouraging use of custom `Robot` classes, which will provide better framework and added functionality (e.g. multiple behaviors).

include/kodlab_mjbots_sdk/mjbots_control_loop.h
ethanmusser: I think you should keep this default implementation (see other discussion in `robot_example.cpp`).

examples/robot_example.cpp
ethanmusser: This should definitely move out of `robot_example`.  I don't think moving to the `JointBase` class makes sense as it is a vector composing one or several `JointBase`'s.  I would just keep this as-is, but stick it in its own class (perhaps in the `joint_base.h` header but outside of the `JointBase` class) to be included as needed.  The implicit conversion and wrapping of `std::shared_ptr` is nice for user interaction.

examples/robot_example.cpp
ethanmusser: If it still provides some useful functionality, just stick it in its own header file to be included as needed.

include/kodlab_mjbots_sdk/common_header.h
```diff
@@ -31,4 +31,20 @@ static void EnableCtrlC() {
   sigaction(SIGINT, &sigIntHandler, NULL);
   CTRL_C_DETECTED = false;
 }
+
+/**
+ * @brief Convert vector to shared pointers using copy constructor
+ * @param vect vector of type T
```
ethanmusser, line 37: Change `vect` to `objs`

src/mjbots_robot_interface.cpp
ethanmusser: Should be consistent everywhere, also `num_joints` is more apt.

examples/robot_example.cpp
ethanmusser: I meant into the same header. My first thought was in something like the common_header but for joints and that is joint_base.h

jdcaporale: I meant into the same header. My first thought was in something like the common_header but for joints and that is joint_base.h

ethanmusser: Moved to joint_header

ethanmusser: Moved to joint_header

ethanmusser:
### Thoughts

- Updated structure is more intuitive.  Using user-defined, `RobotInterface`-derived `Robot` classes is more easily expandable to complex robots with many composed classes and behaviors.  
- See comments in various discussions above.
- Improvements to #16 require the updated infrastructure.  We should get #22 (or both #16 and #22) merged to make these changes.

### TODOs

- [ ] Move LCM mode input addition to separate commit (i.e. merge a separate PR for mode input, then rebase fd4b851c454f5986c7b01015873a63c63686b3dd off of the new commit and rebase a81f6688a616721fc89d664e42eecd35c83525ee off of fd4b851c454f5986c7b01015873a63c63686b3dd)
- [ ] Consolidate `RobotInterface` implementation to `robot_interface.h` only
- [ ] Figure out `CalcTorques` interaction per discussions above
- [ ] Update all usages to `num_joints_` per discussions above
- [ ] General cleanup (i.e., Doxygen comments, format changes per Drake style guide)

src/robot_interface.cpp
jdcaporale: This makes sense. I will do this unless told otherwise. @ShaneRozenLevy?

examples/robot_example.cpp
```diff
@@ -0,0 +1,118 @@
...
+    operator std::vector<std::shared_ptr<joint_type>>() { return v_; }
+};
+
+int main(int argc, char **argv)
+{
+
+    // Setup joints with a std::vector
+    // std::vector<kodlab::mjbots::JointMoteus> joints;
+    // joints.emplace_back(100, 4, 1, 0,   1, 0);
+    // joints.emplace_back(101, 4,-1, 0, 5.0/3.0, 0);
```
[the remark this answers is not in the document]
jdcaporale: Not extraneous, mean to show that you can do it either way

examples/robot_example.cpp
jdcaporale: Done. Also leaving both implementations in the example.

lcm_types/mode_input.lcm
ethanmusser: See #24

examples/robot_example.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include "examples/simple_robot.h"
+
+class SimpleRobotControlLoop : public kodlab::mjbots::MjbotsControlLoop<ManyMotorLog, ModeInput, SimpleRobot>
+{
+    using MjbotsControlLoop::MjbotsControlLoop;
+
+    void CalcTorques() override
+    {
+        robot_->Update();
```
ShaneRozenLevy, line 29: I would add some code here that actually uses the robot to set some torques or atleast a comment expressing that idea as a code pattern.

include/examples/simple_robot.h
```diff
@@ -0,0 +1,27 @@
...
+{
+
+    using kodlab::RobotInterface::RobotInterface;
+
+public:
+    int mode = 0; // Member variables encode whatever added state we need
+    // Set up robot update function for state and torques
+    void Update() override
+    {
+        std::vector<float> torques(num_joints_, 0);
```
ShaneRozenLevy, line 24: I don't think this is what we want in the update as an example for people to pattern match. Maybe give it a leg and use that??

include/kodlab_mjbots_sdk/robot_interface.h
```diff
@@ -0,0 +1,143 @@
...
+#include "kodlab_mjbots_sdk/joint_base.h"
+#include "kodlab_mjbots_sdk/soft_start.h"
+
+namespace kodlab
+{
+    /*!
+     * @brief Sets up a kodlab robot interface class 
+     * 
+     */
+    class RobotInterface {
```
ShaneRozenLevy, line 25: we probably want it to be RobotBase rather than RobotInterface

include/kodlab_mjbots_sdk/mjbots_control_loop.h
ShaneRozenLevy: I think we should not change this. Keep it as: `  virtual void CalcTorques() = 0;`

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -92,33 +112,71 @@ class MjbotsControlLoop : public AbstractRealtimeObject {
...
-  /// initialized inside thread
+  /*!
+   * @brief Construct a new Setup Options object
+   * 
+   * @param options 
+   */
+  void SetupOptions(const ControlLoopOptions &options);
+
+  std::shared_ptr<RobotClass> robot_;     /// ptr to the robot object
+  std::shared_ptr<kodlab::mjbots::MjbotsRobotInterface> mjbots_interface_;   ///ptr to mjbots_interface object, if unique causes issues, also should be initialized inside thread
```
ShaneRozenLevy, line 123: Should this be a unique ptr?

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -189,25 +241,27 @@ void MjbotsControlLoop<log_type, input_type>::Run() {
...
+      mjbots_interface_->ProcessReply();
       prev_msg_duration = message_duration_timer.tac();
       message_duration_timer.tic();
-      robot_->SendCommand();
+      mjbots_interface_->SendCommand();
     }
+
     // Calculate torques and log
-    CalcTorques();
+    CalcTorques();      //TODO should we give full control to the robot_ instead of feeding update thorugh?
```
ShaneRozenLevy, line 251: Lets rename CalcTorques() to Update() and have Update be up to the user what it does. They can call the robot update if they want or just do vanilla calc torques.

include/kodlab_mjbots_sdk/mjbots_robot_interface.h
```diff
@@ -30,32 +30,29 @@ struct RealtimeParams {
   int lcm_cpu = 0;
 };
 
-class MjbotsRobotInterface {
+/*!
+ * @brief A class that allows interaction with the Mjbots Moteus Motor Controllers
+ * 
+ */
+class MjbotsRobotInterface  {
```
ShaneRozenLevy, line 37: Lets rename this class to `MjbotsHardwareInterface`

include/kodlab_mjbots_sdk/robot_interface.h
```diff
@@ -0,0 +1,143 @@
...
+    public:
+        static const int KILL_ROBOT = -1;
+        std::vector< std::shared_ptr<JointBase> > joints; ///the vector of shared_ptrs to joints
```
ShaneRozenLevy, line 28: create getter

include/kodlab_mjbots_sdk/robot_interface.h
```diff
@@ -0,0 +1,143 @@
...
+            // Ensure at compile time that the template is JointBase or a child of JointBase
+            static_assert(std::is_base_of<JointBase, JointDerived>::value);
+
+            // Set up joint vector and joint states
+            for (std::shared_ptr<JointDerived> j : joint_vect){
+                joints.push_back(j); // Copy constructed, uses an implicit upcasting
+                positions_.push_back( j->get_position_reference() );
+                velocities_.push_back( j->get_velocity_reference() );
+                torque_cmd_.push_back( j->get_servo_torque_reference() ); 
+            }
```
ShaneRozenLevy, line 52: Maybe here create a map from joint name to joint_id.

include/kodlab_mjbots_sdk/robot_interface.h
```diff
@@ -0,0 +1,143 @@
...
+        std::vector<float> GetJointTorqueCmd();
+
+        //TODO virtual Attitude GetAttitude()
+
+        /*! 
+         * @brief Get sub-vector of shared_ptr to joint objects via a set of indices
+         * @param joint_indices set of desired joint indices as std::vector of ints
+         * @return a vector shared pointers to the desired joints
+         */
+        std::vector<std::shared_ptr<JointBase>> GetJoints(std::vector<int> joint_indices);
```
ShaneRozenLevy, line 110: Index here by joint_ids rather than by ints.

include/kodlab_mjbots_sdk/robot_interface.h
```diff
@@ -0,0 +1,143 @@
...
+        void SetTorques(std::vector<float> torques);
+
+        /*!
+         * @brief accessor for the torque md, takes into account direction
+         * @return the torque cmd
+         */
+        std::vector<float> GetJointTorqueCmd();
+
+        //TODO virtual Attitude GetAttitude()
+
```
ShaneRozenLevy, line 104: Create a find joint_id_by_name_fatal function

src/robot_interface.cpp
ShaneRozenLevy: seems fine.

ShaneRozenLevy (commented):
seems like a good start and a correct implementation of what we wanted to do. I think as we look towards making more PRs its important to break them into small pieces so they are easier to review.

include/kodlab_mjbots_sdk/mjbots_control_loop.h
ethanmusser: Renamed in 2d28dd28bc0957d92233c7ff3f997166d212ee69.

examples/robot_example.cpp
ethanmusser: I think we should not change this. Keep it as: `  virtual void CalcTorques() = 0;`

_Originally posted by @ShaneRozenLevy in https://github.com/KodlabPenn/kodlab_mjbots_sdk/pull/22#discussion_r931545298_

src/mjbots_robot_interface.cpp
ethanmusser: Resolved in 8d4fd951d65b7cc23aedfa556c2617725146633e.

examples/robot_example.cpp
ethanmusser: I think Shane has a point.  Keep this in `robot_example.cpp` and comment appropriately.  Then make `CalcTorques` (now `Update`) pure virtual in `MjbotsControlLoop`.

include/kodlab_mjbots_sdk/mjbots_control_loop.h
ethanmusser: I've come around to Shane's take so long as a `CalcTorques()` override is provided in `robot_example.cpp` that includes `robot_->Update()` and comments explaining its usage.

ethanmusser: Resolved in 060a657fd0508b266f5dde0c99b0d6b2a72ed7e7.

examples/robot_example.cpp
ethanmusser: `CalcTorqes` made pure virtual in 060a657fd0508b266f5dde0c99b0d6b2a72ed7e7. Documentation of this override should be clarified in a robot_example.cpp documentation update commit.

include/kodlab_mjbots_sdk/mjbots_robot_interface.h
ethanmusser: Renamed in ea7778a442081ea4ef04a3c9f17389b0c9217d31.

include/kodlab_mjbots_sdk/mjbots_control_loop.h
jdcaporale: Everyone seems on board with this.

examples/robot_example.cpp
jdcaporale: LGTM

include/kodlab_mjbots_sdk/mjbots_control_loop.h
jdcaporale: Future plans share the robot interface, its possible that it is useful to be able to share the hardware interface as well, and shared pointers are easier to manage. 
Unique pointer would work for now and could just be passed as a raw pointer, but that is probably bad form.

include/kodlab_mjbots_sdk/mjbots_control_loop.h
jdcaporale: Because of bug in soft_start this will likely break code, but will be fixed before merged.

- [ ] Fix soft start timer - get rid of cycle count

include/kodlab_mjbots_sdk/robot_interface.h
ethanmusser: Renamed in 23bee8968f2a74e9649946cb098f472349cbd7ac.

include/kodlab_mjbots_sdk/robot_interface.h
jdcaporale: Currently public, do we need a getter here as well?

ethanmusser: I don't think we need a getter while this is public.  `joints` could be made private and have access limited to an accessor.

include/kodlab_mjbots_sdk/robot_interface.h
jdcaporale: I think that is a good idea. I'm moving it to an issue (#28)  to be done in it's own PR or with the joint base name  PR (#27).

include/kodlab_mjbots_sdk/robot_interface.h
jdcaporale: I think this should be handled by maps/typedef as well. In particular, so that we can get the joints by CAN_ID or INDEX with INDEX being of type Joint_Index_t or something like that. Related to #28.

include/kodlab_mjbots_sdk/robot_interface.h
jdcaporale: What exactly do you mean by fatal? Should we handle the id error and CTRL-C? Allow the error to kill the program? Or some other jargon I don't know? 
Also, we will be making a map as documented in #28 and this will be handled there instead.

jdcaporale:
Naming convention changes LGTM and new update too (last five commits)

include/examples/simple_robot.h
jdcaporale: Will add better example code

jdcaporale: #36

PR #26, opened 2022-08-02
adds yaml-cpp submodule for config files

jdcaporale:
Adds a YAML submodule from [yaml-cpp](https://github.com/jbeder/yaml-cpp ). Currently not used in kodlab examples, so not included in `examples/CMakeLists.txt` as a dependency. Useful for writing and editing robot configuration files.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "thirdparty/real_time_tools"]
 	path = thirdparty/real_time_tools
 	url = git@github.com:KodlabPenn/real_time_tools.git
+[submodule "thirdparty/yaml-cpp"]
+	path = thirdparty/yaml-cpp
+	url = git@github.com:jbeder/yaml-cpp.git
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "thirdparty/real_time_tools"]
 	path = thirdparty/real_time_tools
 	url = git@github.com:KodlabPenn/real_time_tools.git
+[submodule "thirdparty/yaml-cpp"]
```
ethanmusser, line 4: This submodule is not currently locked to a specific branch in `jbeder/yaml-cpp`.  Since this is actively under development, having the latest version may be desirable, but this also introduces some instability.

ethanmusser (changes requested):
Yaml-cpp needs to be release-locked, but otherwise LGTM.

.gitmodules
ethanmusser: Upon further review, the changes to the `yaml-cpp` `master` branch seem stable enough, and provide useful features.

ethanmusser (approved):
LGTM

.gitmodules
jdcaporale: This was actually set to the latest release branch tagged `yaml-cpp-0.7.0`via checking out that commit in particular. We are not on `master` since that is a live branch. We can update and receive those changes when they release again or if we find any bugs that require it.

PR #33, opened 2022-08-02
Doxyfile and assets

ethanmusser:
Our commenting follows the [Doxygen](https://doxygen.nl/) convention, which allows for automatic documentation generation.  This PR adds a `Doxyfile` for generating the documentation.  With doxygen installed locally, running
```
doxygen Doxyfile
```
will produce HTML documentation files in the `docs/html` directory.  This file is configured to produce documentation in a style similar to the [Robomechanics Lab Quad-SDK](https://robomechanics.github.io/quad-sdk/index.html) documentation. A GitHub Actions workflow for building the documentation and deploying it to GitHub Pages will be included in a seperate PR. 

### Questions

> We can include a nicer looking Doxygen format by setting the `HTML_STYLESHEET` value.  [Doxygen Awesome](https://jothepro.github.io/doxygen-awesome-css/) looks pretty good, we would just have to include it as a submodule.  Should that be included in this PR?

This will be saved for a later PR.

> Setting `GENERATE_LATEX` to `YES` will also generate a LaTeX file output of the documentation.  Do we want this?

Enabled in eade872c7bd1e871c8cced4235deaf7b6846c2ae.

> Python files can be included in the documentation as well, but they clutter the lists.  Do we want them? 

Python classes are enabled, source files for both C++ and Python do not have documentation pages.

> Enabling `USE_MATHJAX` produces inline math rendering.  Is this desireable?  Notice in the `IMUData` example that the math text does not render.

Enabled in 812ef8abdbf476cba91f9190401588026a72fad3.

### Documentation Images

| Main Page   | Class Page  |
|    :----:   |    :----:   |
| doxygen-main-page | doxygen-class-page |

### Example Output

See the attached compressed directories for HTML and LaTeX generated documentation.
[kodlab_mjbots_sdk-docs.zip](https://github.com/KodlabPenn/kodlab_mjbots_sdk/files/9247205/kodlab_mjbots_sdk-docs.zip)


jdcaporale:
> ### Questions
> * In `Doxyfile`:
>   
>   * We can include a nicer looking Doxygen format by setting the `HTML_STYLESHEET` value.  [Doxygen Awesome] (https://jothepro.github.io/doxygen-awesome-css/) looks pretty good, we would just have to include it as a submodule.  Should that be included in this PR?

No we can adjust style in a separate PR if this is working.

>   * Setting `GENERATE_LATEX` to `YES` will also generate a LaTeX file output of the documentation.  Do we want this?

Yes I think this makes sense

>   * `EXCLUDE` dictates files and directories to be omitted.

This is essentially gitignore?

>   * Python files can be included in the documentation as well, but they clutter the lists.  Do we want them?

I don't know how much they clutter things but I think they would be nice to have especially useful utils

>   * Enabling `USE_MATHJAX` produces inline math rendering.  Is this desireable?  Notice in the `IMUData` example that the math text does not render.

I think so, we do math often enough and it would be nice if we could do it easily

ethanmusser:
> > ### Questions
> > 
> > * In `Doxyfile`:
> >   
> >   * We can include a nicer looking Doxygen format by setting the `HTML_STYLESHEET` value.  [Doxygen Awesome] (https://jothepro.github.io/doxygen-awesome-css/) looks pretty good, we would just have to include it as a submodule.  Should that be included in this PR?
> 
> No we can adjust style in a separate PR if this is working.
> 
> > * Setting `GENERATE_LATEX` to `YES` will also generate a LaTeX file output of the documentation.  Do we want this?
> 
> Yes I think this makes sense
> 
> > * `EXCLUDE` dictates files and directories to be omitted.
> 
> This is essentially gitignore?
> 
> > * Python files can be included in the documentation as well, but they clutter the lists.  Do we want them?
> 
> I don't know how much they clutter things but I think they would be nice to have especially useful utils
> 
> > * Enabling `USE_MATHJAX` produces inline math rendering.  Is this desireable?  Notice in the `IMUData` example that the math text does not render.
> 
> I think so, we do math often enough and it would be nice if we could do it easily

These are addressed in 812ef8abdbf476cba91f9190401588026a72fad3 and eade872c7bd1e871c8cced4235deaf7b6846c2ae.

jdcaporale (approved):
LGTM

PR #34, opened 2022-08-02
Doxygen action workflow

ethanmusser:
This action will build the Doxygen documentation and publish the results to the repository's GitHub pages page. The action depends on having a `Doxyfile` present, and would interface with that provided in #33.

### GitHub Pages Deployment

- An [SSH deploy key](https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-create-ssh-deploy-key) or [`GITHUB_TOKEN`](https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-first-deployment-with-github_token) must be configured to deploy to GitHub pages via the `peaceiris/actions-gh-pagesaction` action used in this workflow.  This is configured in the repository settings.
- The documentation will build into and deploy from the `gh-pages` branch.

### DIFF
--- a/.github/workflows/doxygen-action.yml
+++ b/.github/workflows/doxygen-action.yml
@@ -1,0 +1,27 @@
+name: Doxygen Documentation
+
+on:
+  push:
+    branches:
+      - master
+  page_build:
+
+jobs:
+  doxygen:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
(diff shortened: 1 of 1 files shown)


jdcaporale:
Looking into it I was worried about implementing this. Someone suggested making a separate user also for the TOKEN too. It didn't seem this trivial, but maybe it is. Let's meet up and talk about this one. 

I would love for this to just work.

PR #39, opened 2022-08-04
Abstract behavior class and example

ethanmusser:
An abstract class for implementing `RobotBase`-derived robot behaviors.

- Provides an abstract `Behavior` class in the `kodlab` namespace to be used to implement robot behaviors.  These follow the pattern of `Init`, `Begin`, `Update`, and `Stop`, as well as providing a virtual function to permit behavior switching.
- Provides an abstract `IOBehavior` class in the `kodlab` namespace which inherits `Behavior` and provides additional input and output LCM functionality.
- Provides a template `OffBehavior` class in the `kodlab` namespace that is derived from `Behavior`.  This class is templated for a `RobotBase`-derived robot class, and simply commands zero torques to all robot joints inside the `Update` method.
- Provides a `SimpleBehavior` class which demonstrates the intended implementation of `Behavior`-derived classes.
- Provides a `SimpleIOBehavior` class which demonstrates usage of the behavior-level LCM inputs and outputs to execute position control on a joint.

Note that this PR is 1/3 in the behavior implementation series.


ShaneRozenLevy:
How does logging work with behaviors or are we moving to just one update function that also adds vars to log?

README.md
```diff
@@ -111,6 +111,15 @@ class MyController : public MjbotsControlLoop<LcmLog, LcmInput, MyRobot>
 Refer to `include/examples/simple_robot.h` for a sample robot class and 
 `examples/robot_example.cpp` for a usage example.
 
+## Behaviors
+The `Behavior` abstract class can be derived by the user and used to define 
+custom behaviors.  The abstract class includes functions for behavior 
+initialization, startup, updating, and stopping, as well as methods for
```
ShaneRozenLevy, line 117: Provide guidelines on how to use. What do you need to do when you inherit from the behavior base?

include/examples/simple_behavior.h
```diff
@@ -0,0 +1,192 @@
...
+   *          occur, e.g., setting up pointers. In this example, the desired
+   *          torques are initialized to zero.
+   */
+  void Init() override
+  {
+    // Initializes desired torques (for user safety, these are set to zero)
+    desired_torques_ = std::vector<float>(robot_->joints.size(), 0);
+
+    // Indicate that behavior has been initialized
+    set_initialized();
```
ShaneRozenLevy, line 80: This is bad design. The user should not need to call set_initialized() as all behaviors's Init will call set_initialized(). If we want to consider that initialization can fail, we can return a bool, or have the function that call Init, then call set_initialized(). If initialization can fail, what does that mean? Should we just end the program?

include/examples/simple_behavior.h
```diff
@@ -0,0 +1,192 @@
...
+   * @details This code runs a single time when the behavior starts up. This is
+   *          a good place to read in the initial robot state and set the
+   *          appropriate variables to begin the behavior. The previously active
+   *          behavior is passed as an argument, permitting different
+   *          functionalities for different prior behaviors. In this example,
+   *          the behavior is marked as active and the initial robot state is
+   *          read.
+   * @param prev_behavior the previously active behavior
+   */
+  void Begin(std::unique_ptr<kodlab::Behavior<SimpleRobot>> &prev_behavior) override
```
ShaneRozenLevy, line 95: Don't use a reference to a unique ptr. Why not use a reference to the data stored in the ptr?

include/examples/simple_behavior.h
```diff
@@ -0,0 +1,192 @@
...
+  /**
+   * @brief Initializes the behavior
+   * @details This is where initialization code, which runs when a behavior is
+   *          first added to a behavior manager, should be put. This is where
+   *          initialization that did not take place in the constructor should
+   *          occur, e.g., setting up pointers. In this example, the desired
+   *          torques are initialized to zero.
+   */
+  void Init() override
+  {
```
ShaneRozenLevy, line 75: nit: brackets

include/examples/simple_behavior.h
```diff
@@ -0,0 +1,192 @@
...
+            std::vector<float>(robot_->joints.size(), SLOW_TORQUE);
+
+        // Set behavior to inactive if average velocity is below a threshold
+        float avg_velocity =
+            std::accumulate(joint_velocities_.begin(),
+                            joint_velocities_.end(),
+                            0.0) / static_cast<float>(joint_velocities_.size());
+        if (avg_velocity <= 0.1)
+        {
+          set_inactive();
```
ShaneRozenLevy, line 140: The behavior should not set itself inactive, the behavior manager should do that.

include/examples/simple_behavior.h
ShaneRozenLevy, line 95: How does going back to a behavior work?

include/examples/simple_behavior.h
```diff
@@ -0,0 +1,192 @@
...
+   * @brief Stops the behavior
+   * @details This code is run when a behavior is directed to stop running by
+   *          the behavior manager. This should trigger a shutdown process for
+   *          the behavior, e.g. slowing the motor speed. The upcoming behavior
+   *          is passed as an argument to permit different shutdown procedures
+   *          for different on-deck behaviors. In this example, the internal
+   *          state is modified to indicate the behavior should stop.
+   * @param next_behavior the upcoming behavior
+   */
+  void Stop(std::unique_ptr<kodlab::Behavior<SimpleRobot>> &next_behavior) override
```
ShaneRozenLevy, line 160: Will stop be called once or multiple times?

include/examples/simple_behavior.h
```diff
@@ -0,0 +1,192 @@
...
+   *          behavior is prepared to terminate and transition to the next
+   *          behavior. In its simplest form, this function simply accesses
+   *          the internal behavior active flag. This function can be used as a
+   *          mode switch guard.
+   * @param next_behavior the upcoming behavior
+   * @return true if behavior is ready to transition, false otherwise
+   */
+  bool ReadyToSwitch(std::unique_ptr<kodlab::Behavior<SimpleRobot>> &next_behavior) override
+  {
+    return !Running();
```
ShaneRozenLevy, line 188: Add the check on avg velocity here.

README.md
```diff
@@ -111,6 +111,15 @@ class MyController : public MjbotsControlLoop<LcmLog, LcmInput, MyRobot>
 Refer to `include/examples/simple_robot.h` for a sample robot class and 
 `examples/robot_example.cpp` for a usage example.
 
+## Behaviors
```
ShaneRozenLevy, line 114: Add some documentation on how to create the derived class.

ethanmusser:
Yep current implementation has a single logging function called during the control loop update (see `behavior_robot_example.cpp` in #41 for specifics).  I should probably include a custom log which keeps track of the behanvior in this example.

include/kodlab_mjbots_sdk/behavior.h
```diff
@@ -0,0 +1,184 @@
...
+  /**
+   * @brief Behavior name, default is "OFF"
+   */
+  std::string name_;
+
+  /**
+   * @brief Initialization state of this behavior
+   * @note Must be updated appropriately for integration with BehaviorManager
+   */
+  bool initialized_ = false;
```
ShaneRozenLevy, line 49: both initialized and active should be managed by the behavior manager and not the behavior itself.

include/kodlab_mjbots_sdk/behavior.h
```diff
@@ -0,0 +1,184 @@
...
+   */
+  bool active_ = false;
+
+public:
+  /**
+   * @brief Construct a Behavior object
+   * @param robot robot behavior is executing on
+   * @param name[optional] behavior name
+   */
+  Behavior(std::shared_ptr<Robot> robot, std::string name = "EMPTY")
```
ShaneRozenLevy, line 62: use "" instead of "EMPTY"

include/kodlab_mjbots_sdk/behavior.h
```diff
@@ -0,0 +1,184 @@
...
+      : robot_(robot), name_(std::move(name)) {}
+
+  /**
+   * @brief Destructor
+   */
+  virtual ~Behavior() {}
```
ShaneRozenLevy, line 68: Whats wrong with the default Destructor?

jdcaporale:
> How does logging work with behaviors or are we moving to just one update function that also adds vars to log?

We did talk about this. The eventual plan might be that logging inputs  (and more aspirationally the logs themselves) would be handled on a per behavior basis. Currently the BehaviorControlLoop still handles this itself meaning the behaviors all share one logging struct and one input struct. I think different inputs is easy to implement, different outputs requires more coordination from the python script running on the users computer. I'll write the ideas I have in an issue and have you all check it out.

ShaneRozenLevy:
> > How does logging work with behaviors or are we moving to just one update function that also adds vars to log?
> 
> We did talk about this. The eventual plan might be that logging inputs (and more aspirationally the logs themselves) would be handled on a per behavior basis. Currently the BehaviorControlLoop still handles this itself meaning the behaviors all share one logging struct and one input struct. I think different inputs is easy to implement, different outputs requires more coordination from the python script running on the users computer. I'll write the ideas I have in an issue and have you all check it out.

I guess I think its fine to only have 2 LCM types per BehaviorControlLoop (input and output). I just think you need to decide how to connect those into the running behavior.

ShaneRozenLevy (changes requested):
This seems like you are on the correct track, but I have some qualms about some of the details related to LCM IO, and how active and initialized work.

ShaneRozenLevy:
For this set of PR, lets merge them in 1 at a time letting me review pr 2 and 3 once pr 1 is merged into master and master is merged into pr 2, etc.

ShaneRozenLevy:
Follow up on the logging, if a behavior is not the active behavior does it get the input messages or just the most recent one when init, or it doesn't get any at all until its active. I think we should put some serious thought into the input. Output is simpler since we just need to give the behavior access to the output log data message object.

include/kodlab_mjbots_sdk/behavior.h
ethanmusser: This fixed an issue when stopping the control loop that some memory was freed and then accessed.  It didn't affect performance, but I figured it was worth resolving.

It has to do with deleting derived class instances through pointers to their base.  Check out the 'Virtual Destructors' section of [this link](https://en.cppreference.com/w/cpp/language/destructor).

include/examples/simple_behavior.h
ethanmusser: Stop is only called once.  The behavior switch sequence calls `Stop` on the running behavior, then  continues calling `Update` I'm that behavior until it's `ReadyToSwitch` function returns `true`, then switches to the new behavior and calls its `Begin` method.

include/examples/simple_behavior.h
ethanmusser: A behavior is initialized (`Init` is called) during behavior manager initialization (before the control loop begins).  When a behavior becomes active, it's `Begin` is called and then `Update` is called with each control loop update.  Once the behavior's `Stop` is called, the  `Update` loop continues until that behavior's `ReadyToSwitch` (which is polled by the behavior manager) returns `true`.  

Returning to a behavior simply repeats the `Begin` -> `Update` -> `Stop` sequence again.

include/examples/simple_behavior.h
ethanmusser: Somehow I got it in my head that Google style had next-line brackets ..

include/kodlab_mjbots_sdk/behavior.h
ethanmusser: Resolved in 222fdc0bf567682072e4d3765d7dd699569d8554.

include/kodlab_mjbots_sdk/behavior.h
ethanmusser: Changed in 269a9c23597e808ea62a1439b36dc483391a8076.

ethanmusser: Remedied in 269a9c23597e808ea62a1439b36dc483391a8076.

include/examples/simple_behavior.h
ethanmusser: Changed in 3c89ff3de8b8d2cd4de29d27e34cc7a88454cb86.

include/examples/simple_behavior.h
ethanmusser: Removed in c9b815e2ebf6e107820b8e28ad824ffe0d777fbe.

include/examples/simple_behavior.h
ethanmusser: This was implemented in 269a9c23597e808ea62a1439b36dc483391a8076.  Let me know what you think.

include/kodlab_mjbots_sdk/behavior.h
jdcaporale: This makes the destructor virtual which is necessary when you plan to use upcasted pointers.

jdcaporale:
> Follow up on the logging, if a behavior is not the active behavior does it get the input messages or just the most recent one when init, or it doesn't get any at all until its active. I think we should put some serious thought into the input. Output is simpler since we just need to give the behavior access to the output log data message object.

I think it only gets input while active .

To keep the remote system that is communicating in sync with what behavior is active, I would first add active_behavior to the output and later perhaps have a separate message that is Behavior/BehaviorManager specific (maybe just integer indices and bools, or even names if we have the bandwidth).

include/examples/simple_behavior.h
ethanmusser: Switched in d5a5f80f72d7c6a0b4102fe4306e32c69a184078.

include/examples/simple_behavior.h
ShaneRozenLevy: Returning a bool sounds good.

include/kodlab_mjbots_sdk/behavior.h
ShaneRozenLevy: Thats good to know. Can you make an issue to propagate this to other inherited classes. Might fix our segfault

ShaneRozenLevy:
Code looks good. Let me know if you want to chat re IO, but I think to make this useful we will want it. I would like a plan (does not have to be implemented) before we merge in this PR.

jdcaporale:
I'll work with Ethan to make a proper issue with some goals. Likely a band-aid in this PR, but next steps for a different PR.

ethanmusser:
I'm gonna work out a rough solution tomorrow.  Current plan is a single input and output, with the input passed down to the active behavior, and the output maintained at the control loop level, with an additional log field for the selected behavior index.  I think the latter is fine since logging is (at least with our twist stuff) exclusively used for state.  

Long-term, we're definitely thinking individualized inputs for each behavior (if desired), but I agree that a proper solution can come in a later PR.  Will get this issue written tomorrow as well.

include/kodlab_mjbots_sdk/behavior.h
ethanmusser: Will do!

ethanmusser:
The `LcmSubscriber` modifications provided in #47 should permit behavior-specific inputs via a single `LCM` object.  I think that a single, contorl-loop-level log is sufficient, as most of logging is the state and selected behavior.  I can try to bring logging down to behaviors if this is desired, but I just can't foresee a use for it.

jdcaporale:
> The `LcmSubscriber` modifications provided in #47 should permit behavior-specific inputs via a single `LCM` object. `Behavior` will now inherit `LcmMessageHandler`, and will have a template for `InputClass`. The behaviors will be passed a pointer/reference to the `LcmSubscriber` in their constructors, where the subscription will be added via the `LcmSubscriber::AddSubscription` method.
> 
> I think that a single, control-loop-level log is sufficient, as most of logging is the robot state and selected behavior. I can try to bring logging down to behaviors if this is desired, but I just can't envision a good use for it. Also, I don't want to force people (i.e., me and Diego) to have to write a `PrepareLog` override in every single behavior when they will all be identical.

We might want to structure it the same way, i.e. you can have a seperate publish per behavior, but the control loop has one too. It should be much easier to do that than it was to do Input, I have some of the code written already.

ShaneRozenLevy:
So I see how #47 allows for multiple inputs, i.e. each behavior has its own message handling object, but how does that fix the logging issue. I think its fine for us to only have ONE output log per robot, but the behavior needs to have access to that log.

jdcaporale:
> 

My take is that for this PR the assumption is that the Log is purely written in the control loop same as before. So if we wanted access to behavior updated variables they would need to travel up to there, perhaps in the derived robot class or through behavior manager.
If not though, I think it is be relatively trivial to pass along to the behavior manager a pointer to the lcm object therefore giving the manager the ability to publish to the lcm object from the current behavior.

ethanmusser:
@jdcaporale that is essentially my vision and would work more-or-less in the same way that #47 does.  Doing it this way, the user can decide where they want to handle logging and pass pointer/reference to some `LcmPublisher` around accordingly.

ethanmusser:
Question re. the above, should the same LCM object handle subscribing and logging?  Or do we want a separate LCM object on a separate thread?

ShaneRozenLevy:
Logging does not require its own thread as far as I can tell.

jdcaporale:
Yea for now logging is part of our main loop and should stay that way.  Not sure how expensive it is/ could be though. Do you recall from your experiments @ShaneRozenLevy how long many_motor_log took to publish?

include/kodlab_mjbots_sdk/behavior.h
ethanmusser: #49

ethanmusser:
PR #50 provides an LCM publisher object.  My intention is that `Behavior` (and all children) will own an `LcmPublisher` object which shares an `lcm::LCM` object with `MjbotsControlLoop`.  This stores internally a message which can be populated and published on every update, similar to how things are done in `MjbotsControlLoop`.  This implementation should permit logging at both the control loop, and the behavior levels, with optional implementation of both.

ethanmusser:
LCM input and output has been implemented on a behavior level via the new `LcmSubscriber`, `LcmMessagehandler`, and `LcmPublisher` objects.  These are integrated into the `IOBehavior` class, which inherits the `Behavior` class and provides the additional LCM functionality.  LCM inputs and subscribing has been tested with the `MjbotsBehaviorLoop`-derived control loop provided by `behavior_robot_example.cpp` in #41.  There are no issues with having multiple input or output LCM channels running simultaneously.

include/examples/simple_behavior.h
```diff
@@ -0,0 +1,192 @@
...
+   * @brief Indicates if behavior is finished and ready to transition
+   * @details This function indicates to the behavior manager whether the
+   *          behavior is prepared to terminate and transition to the next
+   *          behavior. In its simplest form, this function simply accesses
+   *          the internal behavior active flag. This function can be used as a
+   *          mode switch guard.
+   * @param next_behavior the upcoming behavior
+   * @return true if behavior is ready to transition, false otherwise
+   */
+  bool ReadyToSwitch(const kodlab::Behavior<SimpleRobot> &next_behavior) override
```
jdcaporale, line 176: how might a behavior use this next_behavior to make decisions in its current form? 
Would it require a dynamic_cast? 
Or would it only be possible by name?

include/examples/simple_behavior.h
```diff
@@ -0,0 +1,192 @@
+/**
+ * @file simple_behavior.h
```
jdcaporale, line 2: I might expand or chagne the name of this behavior or its class name to include SpinJoints or SpinAllJoints for safety XD
"SpinAllBehavior" for example

include/examples/simple_io_behavior.h
```diff
@@ -0,0 +1,146 @@
...
+#include "kodlab_mjbots_sdk/io_behavior.h"  // kodlab::IOBehavior
+#include "kodlab_mjbots_sdk/log.h"  // logging macros
+#include "examples/simple_robot.h"  // SimpleRobot
+
+/**
+ * @brief Simple behavior that runs a positional PD control loop on the first
+ * joint in a SimpleRobot with gains adjustable via LCM input and motor state
+ * logged via LCM output
+ */
+class SimpleIOBehavior : public kodlab::IOBehavior<SimpleRobot,
```
jdcaporale, line 30: SimpleControlIOBehavior or something
Again just so people know what is being run apriori

jdcaporale (commented):
This is looking really well done. I have a bunch of comments, feel free to move anything that doesn't seem necessary to issues/future PRs or tell me why I'm wrong.

include/kodlab_mjbots_sdk/io_behavior.h
```diff
@@ -0,0 +1,178 @@
...
+
+  /**
+   * @brief Construct an I/O Behavior object
+   * @param robot robot that behavior is executing on
+   * @param subscriber realtime LCM subscriber to monitor for inputs
+   * @param publisher LCM publisher object to publish outputs with
+   * @param subscribe_channel LCM channel to subscribe to
+   * @param name behavior name
+   */
+  IOBehavior(std::shared_ptr<Robot> robot,
```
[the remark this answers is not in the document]
ethanmusser: Its cleaner to pass around `LcmPublisher` objects with the LCM object and channel name already stored internally than to force a new construction every time.  For our purposes with the `MjbotsbehaviorLoop` in #41, the user-facing constructor is the second one.

include/kodlab_mjbots_sdk/behavior.h
```diff
@@ -137,6 +140,27 @@ class Behavior
     return true;
   }
 
+  /**
+   * @brief Processes input data, if available
+   * @details This method should be implemented by child classes wishing to
+   * handle input processing internally. For supported LCM behavior inputs,
+   * please inherit `IOBehavior`.
+   * @note If using a behavior manager, this method is called every update
+   * cycle, and should implement checking to determine if a new input is
```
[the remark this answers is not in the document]
ethanmusser: I was mostly just trying to keep these additional methods available to the user if desired.  There is some world where additional input is desired (event-based or otherwise), and these methods can be overridden to handle those.  

Since we are working exclusively with LCM, all I/O should be done with a `BehaviorIO` child, and these functions likely will not be used otherwise.  That said, I don't want to sacrifice generality.  If we go the direction of making these private and making the classes friends to `BehaviorManager`, then we should go all the way and do that for all of the `Behavior` methods that are called by the `behaviorManager`.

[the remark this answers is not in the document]
jdcaporale: In that case I would just clarify  the should. If they are implementing something event based it is on them to do that if they are just using it as another update function that is there prerogative too.  So I might change it to 
`@note If using behavior manager, this method is called every update cycle after the Update function.`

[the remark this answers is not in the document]
jdcaporale: I dont know if it actually after or before, but we should clarify that

include/kodlab_mjbots_sdk/io_behavior.h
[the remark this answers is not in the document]
jdcaporale: Since the constructors are something that likely will be overwritten, simplifying is nice. I'll check out where we use this constructor in other PRs

include/examples/simple_behavior.h
jdcaporale: SimpleSpinAllBehavior maybe?

include/kodlab_mjbots_sdk/io_behavior.h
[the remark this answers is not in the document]
ethanmusser: That's fair, I'm fine with getting rid of it until we have a need.

include/kodlab_mjbots_sdk/io_behavior.h
```diff
@@ -0,0 +1,178 @@
...
+  /**
+   * @brief Processes input data, if available
+   */
+  void ProcessInput() final {
+    auto input_msg = input_handler_.GetDataIfNew();
+    if (input_msg.has_value()) {
+      ProcessInput(input_msg.value());
+      input_data_ = input_msg.value();
+    }
+  }
```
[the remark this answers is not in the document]
ethanmusser: The SFINAE solution here requires that the virtual method in `Behavior` must have the same template, and this is not the desired behavior since the user may specify input type `VoidLcm` in `Behavior`, but still be making use of the `ProcessInput()` method for non-LCM inputs.  

A better solution would be to have two different, private `ProcessInputImpl` methods defined in `IOBehavior`, and select an implementation in `ProcessInput` based on tag dispatching.

For our purposes, I think the additional overhead of the simple approach is negligible relative to tag dispatching, and better maintains code readability.

include/kodlab_mjbots_sdk/io_behavior.h
```diff
@@ -0,0 +1,178 @@
...
+
+  /**
+   * @brief Processes input data, if available
+   * @details Virtual method to be implemented when utilizing inputs. Called
+   * when new input data is available, this method processes the new data
+   * provided in `input_msg`.
+   * @param input_msg[in] input data to be processed
+   */
+  virtual void ProcessInput(const Input &input_msg) {}
```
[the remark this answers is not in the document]
ethanmusser: Pursuant to the above changes, the compiler should sort out that this method is never called in the `VoidLcm` case.

include/kodlab_mjbots_sdk/off_behavior.h
```diff
@@ -0,0 +1,50 @@
...
+namespace kodlab
+{
+
+/**
+ * @brief Off behavior commanding zero torques to all joints in a robot
+ *        interface
+ * @tparam Robot[optional] `RobotBase`-derived object
+ */
+template<class Robot = kodlab::RobotBase>
+class OffBehavior : public Behavior<Robot>
```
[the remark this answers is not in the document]
ethanmusser: Yes, users are encouraged to derive children of `OffBehavior`, if desired.  Not interested in permitting dreaded diamond configurations at this time.

lcm_types/motor_log.lcm
```diff
@@ -0,0 +1,10 @@
+struct MotorLog
```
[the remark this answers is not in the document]
ethanmusser: Agreed that logs could be broken out by motor.  Not sure if that buys us anything other than an LCM publisher headache though.  Currently, this message is only provided for use in the `SimpleIOBehavior` example LCM I/O behavior, where the data it is logging is actually redundant (but included for demonstration purposes).

lcm_types/pd_gains.lcm
```diff
@@ -0,0 +1,5 @@
+struct PDGains
+{
+  float  kp;
+  float  kd;
+}
```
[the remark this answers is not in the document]
ethanmusser: The folly of copy-and-paste.

include/examples/simple_behavior.h
ethanmusser: In the present form, the user can either access the `Behavior`'s name and use this for selective switching conditions, or attempt to downcast and use the full behavior.  The former is the intended usage, using only default `Behavior` members for making determinations about the incoming behavior.  However, the latter is permitted.

include/kodlab_mjbots_sdk/io_behavior.h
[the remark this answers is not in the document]
ethanmusser: Changed in 7a9ebe09f21125317b19bed88d7395002bbd1a41

include/kodlab_mjbots_sdk/behavior.h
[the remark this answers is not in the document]
ethanmusser: Clarified in 57cf31c6c2d3f617d25dde54873399c802c2f84c

include/kodlab_mjbots_sdk/behavior.h
```diff
@@ -0,0 +1,205 @@
...
+      : robot_(robot), name_(std::move(name)) {}
+
+  /**
+   * @brief Destructor
+   */
+  virtual ~Behavior() {}
+
+  /**
+   * @brief Initialize this behavior for a given robot interface
+   * @note Must set `initialized_` to true for integration with BehaviorManager
```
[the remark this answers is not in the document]
ethanmusser: Fixed in 57cf31c6c2d3f617d25dde54873399c802c2f84c

include/kodlab_mjbots_sdk/behavior.h
```diff
@@ -0,0 +1,205 @@
...
+   */
+  virtual bool Running() { return active_; }
+
+  /**
+   * @brief Check for whether this behavior is prepared to switch to a new
+   *        behavior
+   * @details This function indicates to the behavior manager whether the
+   *          behavior is prepared to terminate and transition to the next
+   *          behavior. In its simplest form, this function accesses the
+   *          internal behavior active flag. This function can be used as a mode
```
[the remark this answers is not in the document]
ethanmusser: Fixed in 57cf31c6c2d3f617d25dde54873399c802c2f84c

include/kodlab_mjbots_sdk/io_behavior.h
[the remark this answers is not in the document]
ethanmusser: Simple version added in 7a9ebe09f21125317b19bed88d7395002bbd1a41

include/kodlab_mjbots_sdk/io_behavior.h
```diff
@@ -0,0 +1,178 @@
...
+
+  /**
+   * @brief Message handler for LCM inputs.
+   */
+  LcmMessageHandler<Input> input_handler_;
+
+  /**
+   * @brief Cache storing the most recent input data.
+   */
+  [[maybe_unused]] Input input_data_;
```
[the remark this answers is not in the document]
ethanmusser: Naming clarified in 7a9ebe09f21125317b19bed88d7395002bbd1a41

include/examples/simple_behavior.h
```diff
@@ -0,0 +1,192 @@
...
+
+    // Indicate that behavior has been initialized and return
+    std::fprintf(stdout, "Initialized %s\n", get_name().c_str());
+    return true;  // indicates successful initialization
+  }
+
+  /**
+   * @brief Starts the behavior
+   * @details This code runs a single time when the behavior starts up. This is
+   *          a good place to read in the initial robot state and set the
```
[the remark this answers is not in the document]
ethanmusser: Changed to "beginning" in e811f79815be7a70c5f8871d115262860aa8ab88 for clarity.

include/examples/simple_behavior.h
```diff
@@ -0,0 +1,192 @@
...
+   *          appropriate variables to begin the behavior. The previously active
+   *          behavior is passed as an argument, permitting different
+   *          functionalities for different prior behaviors. In this example,
+   *          the behavior is marked as active and the initial robot state is
```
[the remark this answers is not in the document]
ethanmusser: Changed in e811f79815be7a70c5f8871d115262860aa8ab88

include/examples/simple_behavior.h
```diff
@@ -0,0 +1,192 @@
...
+   *          read.
```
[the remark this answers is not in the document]
ethanmusser: Changed in e811f79815be7a70c5f8871d115262860aa8ab88

include/examples/simple_behavior.h
ethanmusser: For the time being, this will remain as is, and it is up to the user to handle downcasting if desired.  In the future, we could implement a `get_type` method that returns the behavior type by adapting a CRTP structure for `Behavior`.

include/examples/simple_io_behavior.h
ethanmusser: Changed in f4b04a6f22a7cb91e3c61c238b5b25a3686f1eef

include/examples/simple_behavior.h
ethanmusser: Changed in f4b04a6f22a7cb91e3c61c238b5b25a3686f1eef

lcm_types/pd_gains.lcm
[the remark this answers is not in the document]
ethanmusser: Fixed in e811f79815be7a70c5f8871d115262860aa8ab88

include/kodlab_mjbots_sdk/io_behavior.h
```diff
@@ -0,0 +1,178 @@
...
+  void PublishOutput() {
+    publisher_.Publish();
+  }
+
+  /**
+   * @brief Publish an output message over LCM.
+   * @param output_msg output message to be published
+   */
+  void PublishOutput(const Output &output_msg) {
+    publisher_data_ = *output_msg;
```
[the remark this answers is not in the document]
ethanmusser: Fixed in 7a9ebe09f21125317b19bed88d7395002bbd1a41

include/kodlab_mjbots_sdk/behavior.h
```diff
@@ -0,0 +1,205 @@
...
+                "Robot must have base kodlab::RobotBase.");
+
+protected:
+  /**
+   * @brief Robot that this behavior runs on
+   */
+  std::shared_ptr<Robot> robot_;
+
+  /**
+   * @brief Behavior name, default is "OFF"
```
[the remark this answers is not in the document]
ethanmusser: Fixed in 57cf31c6c2d3f617d25dde54873399c802c2f84c

README.md
ethanmusser: Added in f5235b4a9f2f70179410d9f36c16d79c5d13f04e

README.md
ethanmusser: Added in f5235b4a9f2f70179410d9f36c16d79c5d13f04e

include/examples/simple_io_behavior.h
```diff
@@ -0,0 +1,146 @@
+/**
+ * @file simple_io_behavior.h
```
ShaneRozenLevy: This behavior stuff is def something that you will want to share when you apply to jobs.

include/examples/simple_io_behavior.h
```diff
@@ -0,0 +1,146 @@
...
+
+    // Retrieve current joint positions
+    joint_positions_ = robot_->GetJointPositions();
+    joint_velocities_ = robot_->GetJointVelocities();
+
+    // Compute command torque for joint 0 via PD, set all others to zero
+    desired_torques_ = std::vector<float>(robot_->joints.size(), 0.0);
+    float pos_error = desired_position_ - joint_positions_[0];
+    float vel_error = DESIRED_VELOCITY - joint_velocities_[0];
+    desired_torques_[0] = gains.kp * pos_error + gains.kd * vel_error;
```
ShaneRozenLevy: At some point I should add my/a PID class.

include/examples/simple_io_behavior.h
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  /**
+   * @brief Prepare output log for publishing
+   * @details Populate the fields of a MotorLog message with the controlled
+   * joint's state.
+   * @param output_msg message to be published
+   */
+  void PrepareOutput(MotorLog &output_msg) override {
+    output_msg.timestamp = time_now_;
```
ShaneRozenLevy: Does this log no longer have the timing information?

include/kodlab_mjbots_sdk/behavior.h
```diff
@@ -0,0 +1,205 @@
...
+    return true;
+  }
+
+  /**
+   * @brief Processes input data, if available
+   * @details This method should be implemented by child classes wishing to
+   * handle input processing internally. For supported LCM behavior inputs,
+   * please inherit `IOBehavior`.
+   * @note If using a behavior manager, this method is called every update
+   * cycle, and should implement checking to determine if a new input is
```
ShaneRozenLevy: Who is responsible for checking if a new input is available prior to using the data?

include/kodlab_mjbots_sdk/behavior.h
```diff
@@ -0,0 +1,205 @@
...
+   * available prior to using the data.
+   */
+  virtual void ProcessInput() {}
```
ShaneRozenLevy, line 154: I'm a little confused why you have these functions here even though this is not an IO behavior.

ShaneRozenLevy: reading forward I don't think these should be here.

include/kodlab_mjbots_sdk/io_behavior.h
```diff
@@ -0,0 +1,178 @@
...
+
+  /**
+   * Virtual destructor for clean destruction of child class instances
+   */
+  virtual ~IOBehavior() = default;
+
+  /**
+   * @brief Processes input data, if available
+   */
+  void ProcessInput() final {
```
ShaneRozenLevy, line 77: Lets call this `ThreadSafeProcessInput` or something of the sort.

ShaneRozenLevy (commented):
Code looks like its pretty close to being good. Overall happy with the IO stuff, just have some questions about some virtual functions in behaviors.

include/kodlab_mjbots_sdk/behavior.h
jdcaporale: This is here because the system currently uses the a vector of pointers to behaviors (upcasted), there are more complicated ways to deal with this issue.

include/examples/simple_io_behavior.h
ethanmusser: The `MotorLog` LCM message in particular does not have margin or message duration in it.  Honestly, I don't think the timing included here is particularly useful, but I wanted to demonstrate how a behavior clock could be made and used to log behavior-specific information.  In my opinion, we should treat the control loop clock as ground truth, and pass the timer around to the various classes that will be logging so that they can include a ground truth timestamp in their log messages.

include/kodlab_mjbots_sdk/behavior.h
ethanmusser: Yes as @jdcaporale said, this was the path of least resistance to allow the `BehaviorManager` to maintain a vector of pointers to `Behavior`-derived classes.  

`Behavior` contains all of the definitions for methods that may be called by `BehaviorManager`.  To allow `ProcessInput` and `ProcessOutput` to be called by `BehaviorManager` when they are used in `IOBehavior`, these definitions must also exist here.  

Alternative solutions could be implemented on the `BehaviorManager` side of things:

- `BehaviorManager` could attempts to downcast the active `Behavior` pointer to an `IOBehavior`, and then only call `ProcessInput` and `ProcessOutput` if the downcast was successful.
- An `AbstractBehavior` class could be made which includes all of the necessary method definitions and is a parent to both `Behavior` and `IOBehavior`.  Then, the vector in `BehaviorManager` could hold pointers to `AbstractBehavior`s instead.

include/kodlab_mjbots_sdk/behavior.h
ethanmusser: If this method is being overridden by the user in a derived class, then it is up to the user to handle checking for whatever their input method is.  In our case, using LCM inputs, this method is overridden in `IOBehavior` using the built-in `LcmMessageHandler` functionalities to check for inputs and process new input, if desired.

include/examples/simple_io_behavior.h
ethanmusser: Yes please!

include/examples/simple_io_behavior.h
ethanmusser: Glad you think it's up to par!  I most certainly will.

include/examples/simple_behavior.h
```diff
@@ -0,0 +1,192 @@
+/**
+ * @file simple_behavior.h
+ * @author Ethan J. Musser (emusser@seas.upenn.edu)
+ * @brief Defines a simple behavior class which commands torques to all joints
```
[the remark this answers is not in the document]
ethanmusser: Provided in cf530a6fea48073d5127b2f8702d83d622e79c2d

include/examples/simple_behavior.h
```diff
@@ -0,0 +1,192 @@
...
+   * @param next_behavior the upcoming behavior
+   * @return true if behavior is ready to transition, false otherwise
+   */
+  bool ReadyToSwitch(const kodlab::Behavior<SimpleRobot> &next_behavior) override
+  {
+    // Ready to switch if average velocity is below a threshold
+    float avg_velocity =
+        std::accumulate(joint_velocities_.begin(),
+                        joint_velocities_.end(),
+                        0.0) / static_cast<float>(joint_velocities_.size());
```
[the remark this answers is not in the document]
ethanmusser: Agreed, fixed in cf530a6fea48073d5127b2f8702d83d622e79c2d

include/examples/simple_behavior.h
```diff
@@ -0,0 +1,192 @@
...
+    if (avg_velocity <= 0.1)
+    {
+      std::fprintf(stdout, "%s is no longer active\n", get_name().c_str());
```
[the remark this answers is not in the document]
ethanmusser: Fixed in cf530a6fea48073d5127b2f8702d83d622e79c2d

include/kodlab_mjbots_sdk/io_behavior.h
ethanmusser: Changed in b7200e36cf66022b5a4e999b30092769db396b6c

include/kodlab_mjbots_sdk/behavior.h
ShaneRozenLevy: ok. I think thats reasonable.

ShaneRozenLevy (approved):
ship it

jdcaporale (approved):
LGTM. Thanks!

PR #40, opened 2022-08-04
Behavior manager class

ethanmusser:
An abstract class for implementing `RobotBase`-derived robot behaviors.

- Provides a `BehaviorManager` class in the `kodlab` namespace to maintain one or several `Behavior`-derived behaviors and manage switching between them.
  - `BehaviorManager` is templated for a `RobotBase`-derived robot, as are the behaviors it holds.
  - Stores behaviors internally in a variable-length container and maintains a default behavior at index 0 of this container (by default, this is the `kodlab::OffBehavior`).
  - Provides methods for adding behaviors and switching between them, including waiting to switch until the running behavior is ready.

Note that this PR is 2/3 in the behavior implementation series and depends on the `Behavior` class introduced in #39.

Features to be implemented in future work include:

- Event-based behavior switching.
- Maintaining a timeout behavior that, if the running behavior does not permit switching in within a set period of time, will be switched to.

### DIFF
--- a/include/kodlab_mjbots_sdk/behavior_manager.h
+++ b/include/kodlab_mjbots_sdk/behavior_manager.h
@@ -120,7 +120,7 @@ public:
       {
         behaviors_[next_idx_]->Init();
       }
-      behaviors_[selected_idx]->Stop(behaviors_[next_idx_]); // stop current behavior
+      behaviors_[selected_idx]->Stop(*behaviors_[next_idx_]); // stop current behavior
       switching_behaviors_ = true;  // indicate transition is in progress
     } else
     {
@@ -220,10 +220,10 @@ public:
   {
     behaviors_[selected_idx]->Update();
     if (switching_behaviors_
-        && behaviors_[selected_idx]->ReadyToSwitch(behaviors_[next_idx_]))
+        && behaviors_[selected_idx]->ReadyToSwitch(*behaviors_[next_idx_]))
     {
       behaviors_[selected_idx]->set_inactive();  // mark previous behavior as inactive
-      behaviors_[next_idx_]->Begin(behaviors_[selected_idx]);  // begin new behavior
+      behaviors_[next_idx_]->Begin(*behaviors_[selected_idx]);  // begin new behavior
       behaviors_[next_idx_]->set_active();  // mark new behavior as active
       selected_idx = next_idx_;  // update selected behavior to next
       switching_behaviors_ = false;  // no longer transitioning behaviors
--- a/include/kodlab_mjbots_sdk/behavior_manager.h
+++ b/include/kodlab_mjbots_sdk/behavior_manager.h
@@ -57,7 +57,7 @@ class BehaviorManager {
   const std::string KILL_BEHAVIOR_NAME = "KILL";
 
   /**
-   * @brief Behavior command that turns robot off, always 0
+   * @brief Index for the default behavior (that turns robot off by default)
    */
   const int DEFAULT_IDX = 0;
 
@@ -68,8 +68,8 @@ class BehaviorManager {
 
   /**
    * @brief Behaviors available to the robot
-   * @note The first element of this vector is always `OFF_BEHAVIOR`.  Added
-   *       behaviors are appended to the end.
+   * @note The first element of this vector is always `DEFAULT_BEHAVIOR`
+   * (`OFF_BEHAVIOR` by default).  Added behaviors are appended to the end.
    */
   std::vector<std::unique_ptr<Behavior<Robot>>> behaviors_;
 
@@ -115,7 +115,9 @@ class BehaviorManager {
     if (b->Init()) {
       b->set_initialized();
       return std::move(b);
-    } else { return nullptr; }
+    } else {
+      return nullptr;
+    }
   }
 
  public:
@@ -137,7 +139,7 @@ class BehaviorManager {
    * @brief Set the current behavior
    * @param next_idx index in `behaviors_`
    */
-  virtual void SetBehavior(const int &next_idx) {
+  virtual void SetBehavior(int next_idx) {
     if (next_idx == KILL_ROBOT_IDX) {
       kodlab::CTRL_C_DETECTED = true; // kill robot
       LOG_FATAL("Behavior %d received, killing robot.", KILL_ROBOT_IDX);
@@ -164,8 +166,9 @@ class BehaviorManager {
     if (name == KILL_BEHAVIOR_NAME) { SetBehavior(KILL_ROBOT_IDX); }
     else {
       auto idx = get_behavior_index(name);
-      if (idx) { SetBehavior(idx.value()); }
-      else {
+      if (idx) {
+        SetBehavior(idx.value());
+      } else {
         LOG_WARN("Invalid behavior selected: %s", name.c_str());
       }
     }
@@ -217,8 +220,8 @@ class BehaviorManager {
       name_to_idx_map_.erase(default_name_);
       default_name_ = b->get_name();
       name_to_idx_map_.emplace(std::pair(default_name_, DEFAULT_IDX));
-      names_[0] = default_name_;
-      behaviors_[0] = std::move(b);
+      names_[DEFAULT_IDX] = default_name_;
+      behaviors_[DEFAULT_IDX] = std::move(b);
     } else {
       LOG_WARN(
           "%s behavior could not be initialized and will not be set as default behavior.",
@@ -233,8 +236,8 @@ class BehaviorManager {
     if (behaviors_.size() > 1) {
       behaviors_.erase(behaviors_.begin() + 1, behaviors_.end());
       names_.erase(names_.begin() + 1, names_.end());
-      for (const auto &it : name_to_idx_map_) {
-        if (it.first != default_name_) { name_to_idx_map_.erase(it.first); }
+      for (const auto &el : name_to_idx_map_) {
+        if (el.first != default_name_) { name_to_idx_map_.erase(el.first); }
       }
     }
   }
--- a/include/kodlab_mjbots_sdk/behavior_manager.h
+++ b/include/kodlab_mjbots_sdk/behavior_manager.h
@@ -218,7 +218,9 @@ public:
    */
   virtual void Update()
   {
+    behaviors_[selected_idx]->ProcessInput();
     behaviors_[selected_idx]->Update();
+    behaviors_[selected_idx]->ProcessOutput();
     if (switching_behaviors_
         && behaviors_[selected_idx]->ReadyToSwitch(*behaviors_[next_idx_]))
     {
--- a/include/kodlab_mjbots_sdk/behavior_manager.h
+++ b/include/kodlab_mjbots_sdk/behavior_manager.h
@@ -51,6 +51,11 @@ private:
   static const int KILL_ROBOT_IDX = -1;
 
   /**
+   * @brief Behavior command that turns robot off, always 0
+   */
+  static const int DEFAULT_IDX = 0;
+
+  /**
    * @brief Behaviors available to the robot
    * @note The first element of this vector is always `OFF_BEHAVIOR`.  Added
    *       behaviors are appended to the end.
@@ -83,7 +88,10 @@ public:
    * @brief Construct a behavior manager object
    * @param robot pointer to robot that the behavior is executing on
    */
-  BehaviorManager(std::shared_ptr<Robot> robot) : robot_(robot) {}
+  BehaviorManager(std::shared_ptr<Robot> robot) : robot_(robot)
+  {
+    AddBehavior<OffBehavior<Robot>>(robot_, "OFF");
+  }
 
   /**
    * @brief Destructor
@@ -94,7 +102,31 @@ public:
    * @brief Set the current behavior
    * @param next_idx index in `behaviors_`
    */
-  virtual void SetBehavior(const int &next_idx);
+  virtual void SetBehavior(const int &next_idx)
+  {
+    if (next_idx == KILL_ROBOT_IDX)
+    {
+      kodlab::CTRL_C_DETECTED = true; // kill robot
+      LOG_FATAL("Behavior %d received, killing robot.", KILL_ROBOT_IDX);
+    } else if (next_idx == selected_idx)
+    {
+      LOG_INFO("Behavior %d is already active.", next_idx);
+      return;
+    }
+    else if (next_idx < behaviors_.size() && next_idx >= 0)
+    {
+      next_idx_ = next_idx; // update next behavior
+      if (!behaviors_[next_idx_]->is_initialized())
+      {
+        behaviors_[next_idx_]->Init();
+      }
+      behaviors_[selected_idx]->Stop(behaviors_[next_idx_]); // stop current behavior
+      switching_behaviors_ = true;  // indicate transition is in progress
+    } else
+    {
+      LOG_WARN("Invalid behavior selected: %d", next_idx);
+    }
+  }
 
   /**
    * @brief Add a behavior to the behaviors list and initialize it
@@ -103,18 +135,63 @@ public:
    * @param args behavior constructor arguments
    */
   template<class BehaviorType, typename... ConstructorArgs>
-  void AddBehavior(ConstructorArgs &&... args);
+  void AddBehavior(ConstructorArgs &&... args)
+  {
+    static_assert(std::is_base_of<kodlab::Behavior<Robot>, BehaviorType>::value,
+                  "BehaviorType must be a `kodlab::Behavior`-derived type");
+    behaviors_.emplace_back(std::make_unique<BehaviorType>(args...));
+    names_.push_back(behaviors_.back()->get_name());
+    behaviors_.back()->Init();
+  }
 
   /**
    * @brief Reset the behaviors list back to default
    */
-  void ResetBehaviors();
+  void ResetBehaviors()
+  {
+    if (behaviors_.size() > 1)
+    {
+      behaviors_.erase(behaviors_.begin() + 1, behaviors_.end());
+      names_.erase(names_.begin() + 1, names_.end());
+    }
+  }
 
   /**
    * @brief Update the current behavior
    * @param robot shared pointer to robot behavior is executing on
    */
-  virtual void Update();
+  virtual void Update()
+  {
+    behaviors_[selected_idx]->Update();
+    if (switching_behaviors_
+        && behaviors_[selected_idx]->ReadyToSwitch(behaviors_[next_idx_]))
+    {
+      behaviors_[selected_idx]->set_inactive();  // mark previous behavior as inactive
+      behaviors_[next_idx_]->Begin(behaviors_[selected_idx]);  // begin new behavior
+      behaviors_[next_idx_]->set_active();  // mark new behavior as active
+      selected_idx = next_idx_;  // update selected behavior to next
+      switching_behaviors_ = false;  // no longer transitioning behaviors
+      std::fprintf(stdout,
+                   "[INFO][BehaviorManager] Switched to behavior %d.\n",
+                   selected_idx);
+    }
+  }
+
+  /**
+   * @brief Prints a list of behaviors currently in `behaviors_`
+   */
+  void PrintBehaviorList() const
+  {
+    std::fprintf(stdout, "+--------------------------------+\n");
+    std::fprintf(stdout, "| Control Loop Behaviors         |\n");
+    std::fprintf(stdout, "| ----------------------         |\n");
+    for (unsigned int i = 0; i < names_.size(); i++)
+    {
+      std::fprintf(stdout, "| %3d. %-25s |\n",
+                   i, names_[i].c_str());
+    }
+    std::fprintf(stdout, "+--------------------------------+\n");
+  }
 
   /**
    * @brief Accessor for index of selected behavior
--- a/include/kodlab_mjbots_sdk/behavior_manager.h
+++ b/include/kodlab_mjbots_sdk/behavior_manager.h
@@ -14,13 +14,14 @@
 
 #include <iostream>
 #include <memory>
+#include <unordered_map>
+#include <utility>
 #include "kodlab_mjbots_sdk/robot_base.h"
 #include "kodlab_mjbots_sdk/behavior.h"
 #include "kodlab_mjbots_sdk/off_behavior.h"
 #include "kodlab_mjbots_sdk/log.h"
 
-namespace kodlab
-{
+namespace kodlab {
 
 /**
  * @brief Manager for `Behavior` objects
@@ -34,12 +35,11 @@ namespace kodlab
  * @tparam Robot[optional] derived `kodlab::RobotBase` class
  */
 template<class Robot = kodlab::RobotBase>
-class BehaviorManager
-{
+class BehaviorManager {
   static_assert(std::is_base_of<kodlab::RobotBase, Robot>::value,
                 "Robot must have base kodlab::RobotBase.");
 
-private:
+ private:
   /**
    * @brief Robot that behaviors run on
    */
@@ -48,12 +48,22 @@ private:
   /**
    * @brief Behavior command that kills robot, default -1
    */
-  static const int KILL_ROBOT_IDX = -1;
+  const int KILL_ROBOT_IDX = -1;
+
+  /**
+   * @brief Name of the kill behavior
+   */
+  const std::string KILL_BEHAVIOR_NAME = "KILL";
 
   /**
    * @brief Behavior command that turns robot off, always 0
    */
-  static const int DEFAULT_IDX = 0;
+  const int DEFAULT_IDX = 0;
+
+  /**
+   * @brief Name of the default behavior
+   */
+  std::string default_name_;
 
   /**
    * @brief Behaviors available to the robot
@@ -61,6 +71,11 @@ private:
    *       behaviors are appended to the end.
    */
   std::vector<std::unique_ptr<Behavior<Robot>>> behaviors_;
+
+  /**
+   * @brief Map from behavior name to index in `behaviors_`
+   */
+  std::unordered_map<std::string, int> name_to_idx_map_;
 
   /**
    * @brief Names of behaviors in `behaviors_`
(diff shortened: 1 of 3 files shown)


include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,295 @@
...
+   * @brief Behaviors available to the robot
+   * @note The first element of this vector is always `OFF_BEHAVIOR`.  Added
+   *       behaviors are appended to the end.
+   */
+  std::vector<std::unique_ptr<Behavior<Robot>>> behaviors_;
+
+  /**
+   * @brief Names of behaviors in `behaviors_`
+   */
+  std::vector<std::string> names_;
```
ShaneRozenLevy, line 68: Please add a map from name to index

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,295 @@
...
+  /**
+   * @brief Destructor
+   */
+  virtual ~BehaviorManager() {}
+
+  /**
+   * @brief Set the current behavior
+   * @param next_idx index in `behaviors_`
+   */
+  virtual void SetBehavior(const int &next_idx)
```
ShaneRozenLevy, line 105: provide an overload that uses the map

include/kodlab_mjbots_sdk/behavior_manager.h
ShaneRozenLevy: (unordered map is probably best)

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,295 @@
...
+      next_idx_ = next_idx; // update next behavior
+      if (!behaviors_[next_idx_]->is_initialized())
+      {
+        behaviors_[next_idx_]->Init();
+      }
+      behaviors_[selected_idx]->Stop(*behaviors_[next_idx_]); // stop current behavior
+      switching_behaviors_ = true;  // indicate transition is in progress
+    } else
+    {
+      LOG_WARN("Invalid behavior selected: %d", next_idx);
```
ShaneRozenLevy, line 127: I feel like we should not just warn when this happens.

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,295 @@
...
+   * @param next_idx index in `behaviors_`
+   */
+  virtual void SetBehavior(const int &next_idx)
+  {
+    if (next_idx == KILL_ROBOT_IDX)
+    {
+      kodlab::CTRL_C_DETECTED = true; // kill robot
+      LOG_FATAL("Behavior %d received, killing robot.", KILL_ROBOT_IDX);
+    } else if (next_idx == selected_idx)
+    {
```
ShaneRozenLevy, line 112: nit: brackets

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,295 @@
...
+  /**
+   * @brief Constructs and initializes a behavior
+   * @tparam BehaviorType `kodlab::Behavior`-derived type
+   * @tparam ConstructorArgs constructor arguments
+   * @param args behavior constructor arguments
+   * @return unique pointer to the behavior if initialization was successful,
+   *         nullptr otherwise
+   */
+  template<class BehaviorType, typename... ConstructorArgs>
+  std::unique_ptr<BehaviorType> InitBehavior(ConstructorArgs &&... args)
```
ShaneRozenLevy, line 140: I don't think init behavior is the correct function name. Maybe add behavior?

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,295 @@
...
+  {
+    static_assert(std::is_base_of<kodlab::Behavior<Robot>, BehaviorType>::value,
+                  "BehaviorType must be a `kodlab::Behavior`-derived type");
+    auto b = std::make_unique<BehaviorType>(args...);
+    if (b->Init())
+    {
+      b->set_initialized();
+      return std::move(b);
```
ShaneRozenLevy, line 148: What is the purpose of this function? When do you expect it to be used?

include/kodlab_mjbots_sdk/behavior_manager.h
ShaneRozenLevy: Proposal change to CreateBehavior and make it private.

include/kodlab_mjbots_sdk/behavior_manager.h
ShaneRozenLevy: see above comment

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,295 @@
...
+                   i, names_[i].c_str());
+    }
+    std::fprintf(stream, "+--------------------------------+\n");
+  }
+
+  /**
+   * @brief Accessor for index of selected behavior
+   * @return selected behavior index
+   */
+  [[nodiscard]] unsigned int get_selected_behavior() const
```
ShaneRozenLevy, line 259: should specify index in function name. This makes me think it returns a pointer to the behavior.

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,295 @@
...
+  [[nodiscard]] std::string get_sel_behavior_name() const
+  {
+    return get_behavior_name(selected_idx);
+  }
+
+  /**
+   * @brief Accessor for all behavior names in `behaviors_`
+   * @return vector of names of behaviors
+   */
+  [[nodiscard]] std::vector<std::string> get_behavior_list() const
```
ShaneRozenLevy, line 287: `get_behavior_names.

Current function name makes me think its the vector of behaviors.

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,295 @@
...
+#pragma once
+
+#include <iostream>
+#include <memory>
+#include "kodlab_mjbots_sdk/robot_base.h"
+#include "kodlab_mjbots_sdk/behavior.h"
+#include "kodlab_mjbots_sdk/off_behavior.h"
+#include "kodlab_mjbots_sdk/log.h"
+
+namespace kodlab
```
ShaneRozenLevy, line 22: lets think about making a name space and possibly a separate folder for behavior related code. It will make it easier to share the code you wrote when you apply for jobs.

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,295 @@
...
+      LOG_WARN(
+          "%s behavior could not be initialized and will not be set as default behavior.",
+          b->get_name().c_str());
+    }
+  }
+
+  /**
+   * @brief Reset the behaviors list back to default
+   */
+  void ResetBehaviors()
```
ShaneRozenLevy, line 206: `EraseBehaviors` Reset makes me think the it just sets the behaviors back to their default state rather than removing all of them.

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  /**
+   * @brief Add a behavior to the behaviors list and initialize it
+   * @tparam BehaviorType `kodlab::Behavior`-derived type
+   * @tparam ConstructorArgs constructor arguments
+   * @param args behavior constructor arguments
+   */
+  template<class BehaviorType, typename... ConstructorArgs>
+  void AddBehavior(ConstructorArgs &&... args)
```
ShaneRozenLevy, line 159: Lets return the index for the behavior.

include/kodlab_mjbots_sdk/behavior_manager.h
ShaneRozenLevy, line 105: How do people feel about providing the type of behavior and the checking to make sure the relevant behavior is the correct type? IDK how trivial it would be to code, but I just want to minimize the number of times the code enters the wrong behavior.

ShaneRozenLevy (commented):
Code looks to be off to a good start. A few comments about function names, and I think adding an unorded map from string to index would make it much easier to use.

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: For behaviors, I think that warning is sufficient.  The way that @jdcaporale and I use this involves a simple CLI which prompts for a mode number and publishes it.  It is not uncommon for me to incorrectly enter a mode that is out of range.  In this case, I would prefer just to be warned than have the robot shut off or send the motors into braking.

If you feel differently, perhaps we could add an optional parameter when constructing the `BehaviorManager` to dictate what the robot should do in this circumstance (i.e., warn, return to default behavior, or kill).

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Sounds good to me!  If it's alright with you, I will wait and do this in a separate PR once #40 and #41 are merged since we have already merged #39 (`Behavior` and `IOBehavior`) in the current `kodlab` namespace.

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: This would definitely be safer, but it would require additional, redundant changes to user-defined `ProcessInput` methods to implement in full (e.g., defining a switch with cases for each expected behavior index, then running the appropriate `SetBehavior` template method.

Something that was talked about at some point was implementing a boolean behavior transition matrix, which the user could provide to dictate which inter-behavior transitions are acceptable.  I think this would not be particularly difficult to implement if it was set after all behaviors were added to the manager.  Adding as you go would be a little more tricky.  Either way, I would probably keep this as an optional feature.

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: This function constructs the behavior from the given arguments and calls the behavior's initialization method.  If the initialization method succeeds, the behavior is returned to be added to the behaviors list.  Otherwise, a `nullptr` is returned.

The name change and move to private sound good!

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Sounds good!

include/kodlab_mjbots_sdk/behavior_manager.h
ShaneRozenLevy: Ok, happy to leave the check as is. I think the string based argument will make things much easier to use.

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Provided in d5a6010ee947e62b6bbd122659ed3fcc02ff3224

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Added in d5a6010ee947e62b6bbd122659ed3fcc02ff3224

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Fixed in d5a6010ee947e62b6bbd122659ed3fcc02ff3224

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Changed in d5a6010ee947e62b6bbd122659ed3fcc02ff3224

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Changed in d5a6010ee947e62b6bbd122659ed3fcc02ff3224

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Fixed in d5a6010ee947e62b6bbd122659ed3fcc02ff3224

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Renamed in d5a6010ee947e62b6bbd122659ed3fcc02ff3224

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Renamed in d5a6010ee947e62b6bbd122659ed3fcc02ff3224

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Return value added in d5a6010ee947e62b6bbd122659ed3fcc02ff3224

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: An issue has been opened and assigned (#57).

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Agreed, definitely a good idea!  I will see about the transition matrix in a future issue.

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -83,93 +98,105 @@ class BehaviorManager
...
+      name_to_idx_map_.emplace(std::make_pair(name, idx));
       behaviors_.emplace_back(std::move(b));
-    } else
-    {
+      return idx;
+    } else {
       LOG_WARN(
           "%s behavior could not be initialized and will not be added to behaviors list.",
           b->get_name().c_str());
+      return -1;
```
ShaneRozenLevy, line 199: hmm. Thats probably bad since -1 is the kill behavior index. lets make it an optional int.

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -83,93 +98,105 @@ class BehaviorManager
...
-    auto b = std::make_unique<BehaviorType>(args...);
-    if (b->Init())
-    {
-      b->set_initialized();
-      return std::move(b);
-    } else { return nullptr; }
+  virtual void SetBehavior(const std::string &name) {
+    if (name == KILL_BEHAVIOR_NAME) { SetBehavior(KILL_ROBOT_IDX); }
+    else {
+      auto idx = get_behavior_index(name);
```
ShaneRozenLevy, line 165: lets use std::optional here

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -266,26 +289,38 @@ class BehaviorManager
...
+   * @param name behavior name
+   * @return index of behavior if behavior is in the behavior list, -1 otherwise
+   */
+  [[nodiscard]] int get_behavior_index(const std::string &name) {
+    auto it = name_to_idx_map_.find(name);
+    if (it != name_to_idx_map_.end()) {
+      return it->second;
+    } else {
+      LOG_WARN("%s is not in behavior list.", name.c_str());
+      return -1;
```
ShaneRozenLevy, line 307: std::optional. Other option would be to return the iterator, and have the use check if its equal to the end, but the std::optional is a clearer design

ShaneRozenLevy (commented):
Looking close to good. A few small changes, then I'll look everything over at should be good.

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Haha probably for the best, fixed in 6bc58c22c7bfe4d156e3962308c3fbf73fc639c6

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Agreed `std::optional` is the way.  Added in 6bc58c22c7bfe4d156e3962308c3fbf73fc639c6

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Changed in 6bc58c22c7bfe4d156e3962308c3fbf73fc639c6

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,330 @@
...
+
+  /**
+   * @brief Behavior command that kills robot, default -1
+   */
+  const int KILL_ROBOT_IDX = -1;
+
+  /**
+   * @brief Name of the kill behavior
+   */
+  const std::string KILL_BEHAVIOR_NAME = "KILL";
```
ShaneRozenLevy, line 57: https://www.wired.com/2015/11/null/

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,330 @@
...
+   * @param args behavior constructor arguments
+   * @return unique pointer to the behavior if initialization was successful,
+   *         nullptr otherwise
+   */
+  template<class BehaviorType, typename... ConstructorArgs>
+  std::unique_ptr<BehaviorType> ConstructBehavior(ConstructorArgs &&... args) {
+    static_assert(std::is_base_of<kodlab::Behavior<Robot>, BehaviorType>::value,
+                  "BehaviorType must be a `kodlab::Behavior`-derived type");
+    auto b = std::make_unique<BehaviorType>(args...);
+    if (b->Init()) {
```
ShaneRozenLevy, line 115: In the current paradigm, is init ever called not right after the constructor?

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,330 @@
...
+  /**
+   * @brief Destructor
+   */
+  virtual ~BehaviorManager() = default;
+
+  /**
+   * @brief Set the current behavior
+   * @param next_idx index in `behaviors_`
+   */
+  virtual void SetBehavior(const int &next_idx) {
```
ShaneRozenLevy, line 140: Don't pass int by reference: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-conventional

ShaneRozenLevy (approved):
LGTM only blocker is passing int by reference.

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,330 @@
...
+
+  /**
+   * @brief Set the current behavior
+   * @param name name of behavior in `behaviors_`
+   */
+  virtual void SetBehavior(const std::string &name) {
+    if (name == KILL_BEHAVIOR_NAME) { SetBehavior(KILL_ROBOT_IDX); }
+    else {
+      auto idx = get_behavior_index(name);
+      if (idx) { SetBehavior(idx.value()); }
```
jdcaporale, line 167: This is worse as far as formatting than the previous IMO.

jdcaporale, line 167:
```suggestion
      if (idx) {
        SetBehavior(idx.value());
      }
```

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,330 @@
...
+   */
+  template<class BehaviorType, typename... ConstructorArgs>
+  std::unique_ptr<BehaviorType> ConstructBehavior(ConstructorArgs &&... args) {
+    static_assert(std::is_base_of<kodlab::Behavior<Robot>, BehaviorType>::value,
+                  "BehaviorType must be a `kodlab::Behavior`-derived type");
+    auto b = std::make_unique<BehaviorType>(args...);
+    if (b->Init()) {
+      b->set_initialized();
+      return std::move(b);
+    } else { return nullptr; }
```
jdcaporale, line 118:
```suggestion
    } else {
      return nullptr;
    }
```

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,330 @@
...
+  void SetDefaultBehavior(ConstructorArgs &&... args) {
+    static_assert(std::is_base_of<kodlab::Behavior<Robot>, BehaviorType>::value,
+                  "BehaviorType must be a `kodlab::Behavior`-derived type");
+    auto b = ConstructBehavior<BehaviorType>(args...);
+    if (b) {
+      name_to_idx_map_.erase(default_name_);
+      default_name_ = b->get_name();
+      name_to_idx_map_.emplace(std::pair(default_name_, DEFAULT_IDX));
+      names_[0] = default_name_;
+      behaviors_[0] = std::move(b);
```
jdcaporale, line 221: shouldnt these use `DEFAULT_IDX`?

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,330 @@
...
+    } else {
+      LOG_WARN(
+          "%s behavior could not be initialized and will not be set as default behavior.",
+          b->get_name().c_str());
```
jdcaporale, line 225: If the behavior fails to construct, would it have a get_name member?

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,330 @@
...
+
+  /**
+   * @brief Map from behavior name to index in `behaviors_`
+   */
+  std::unordered_map<std::string, int> name_to_idx_map_;
+
+  /**
+   * @brief Names of behaviors in `behaviors_`
+   */
+  std::vector<std::string> names_;
```
jdcaporale, line 84: I'm torn. My gut says names_ that we shouldn't keep two vectors that match index the way names_ and behaviors_ do. 
Instead just query behaviors_.get_name().

It is nice for convenience, but a possible vector for errors and bugs. 
Most of the places that use names switch effortlessly. 
The member function get_behavior_names() will need a for loop or transform or something. 

The map is a useful thing to go the other way though so happy have that even though it also has a similar issue.

jdcaporale (approved):
Couple possible bugs and some philosophical and formatting changes, but all small so LGTM after you deal with that.

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: No, currently `Init` is only called right after the constructor.  `Init` called outside the constructor as the [style guide](https://drake.mit.edu/styleguide/cppguide.html#Doing_Work_in_Constructors) discourages calling virtual methods in constructors (for good reason).  This separation is also maintained in case a `BehaviorManager` child wishes to call `Init` for all contained behaviors at the same time.

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: A `false` return value from `ConstructBehavior` actually indicates that the behavior failed to initialize (i.e., the `Init` method returned `false`).  If the behavior fails to construct due to something in runtime, the program will terminate with a segmentation fault.  However, all other constructor issues should be caught at compile time.

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: This was added to avoid repeated calls to `Behavior::get_name`, particularly the looping in `get_behavior_named`.  I could go either way.

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Fixed in e1e7ee036a97fdee4c755d57904d3f9dc82879e4

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,330 @@
...
+   */
+  const int KILL_ROBOT_IDX = -1;
+
+  /**
+   * @brief Name of the kill behavior
+   */
+  const std::string KILL_BEHAVIOR_NAME = "KILL";
+
+  /**
+   * @brief Behavior command that turns robot off, always 0
```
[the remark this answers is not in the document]
ethanmusser: changed in 57a65848a1b429d69bd735eddc489b60878e210b

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,330 @@
...
+  const int DEFAULT_IDX = 0;
+
+  /**
+   * @brief Name of the default behavior
+   */
+  std::string default_name_;
+
+  /**
+   * @brief Behaviors available to the robot
+   * @note The first element of this vector is always `OFF_BEHAVIOR`.  Added
```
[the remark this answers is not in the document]
ethanmusser: Clarified in 57a65848a1b429d69bd735eddc489b60878e210b

include/kodlab_mjbots_sdk/behavior_manager.h
```diff
@@ -0,0 +1,330 @@
...
+  }
+
+  /**
+   * @brief Reset the behaviors list back to default
+   */
+  void EraseBehaviors() {
+    if (behaviors_.size() > 1) {
+      behaviors_.erase(behaviors_.begin() + 1, behaviors_.end());
+      names_.erase(names_.begin() + 1, names_.end());
+      for (const auto &it : name_to_idx_map_) {
```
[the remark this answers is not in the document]
ethanmusser: Changed to `el` (element) in 57a65848a1b429d69bd735eddc489b60878e210b.

include/kodlab_mjbots_sdk/behavior_manager.h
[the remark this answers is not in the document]
ethanmusser: Formatted per style guide in 57a65848a1b429d69bd735eddc489b60878e210b

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Formatted per style guide in 57a65848a1b429d69bd735eddc489b60878e210b

include/kodlab_mjbots_sdk/behavior_manager.h
ethanmusser: Yes they should, fixed in 57a65848a1b429d69bd735eddc489b60878e210b

PR #42, opened 2022-08-04
Adds virtual intitialization method to MjbotsControlLoop

### LINKED ISSUE #23: MjbotsControlLoop should have init member function
Currently any one off setup needs to be custom made into a constructor or put behind some sort of initial_call flag in CalcTorques(). MjbotsControlLoop should probably have a non-pure virtual function Init(), Begin(),  or equivalent.

ethanmusser:
Adds a virtual intialization function to `MjbotsControlLoop` for initializing members of derived classes.  In its present state, `MjbotsControlLoop`-derivatives must write their own constructors (of which there are several to duplicate from `MjbotsControlLoop`) in order to intialize their members.  This closes #23.


jdcaporale (approved):
LGTM

PR #44, opened 2022-08-04
Moves examples to kodlab::examples namespace

### LINKED ISSUE #32: Examples namespace
Currently examples live in the top-level namespace.  I suggest we move these to an `examples` namespace.

ethanmusser:
Currently examples clutter the top-level namespace.  This PR moves all the current SDK examples to the `kodlab::examples` namespace.  Closes #32.

### DIFF
--- a/examples/imu_example.cpp
+++ b/examples/imu_example.cpp
@@ -15,6 +15,9 @@
 #include "kodlab_mjbots_sdk/imu_data.h"
 #include "IMULog.hpp"
 
+namespace kodlab::examples
+{
+
 class AttitudeExample : public kodlab::mjbots::MjbotsControlLoop<IMULog,
                                                                  VoidLcm>
 {
@@ -92,6 +95,9 @@ private:
 
 };
 
+} // kodlab::examples
+
+
 int main(int argc, char **argv)
 {
   // Setup Joints
@@ -119,6 +125,7 @@ int main(int argc, char **argv)
   options.imu_world_offset_deg.yaw = 0;
 
   // Create Control Loop
+  using kodlab::examples::AttitudeExample;
   AttitudeExample control_loop(joints, options);
 
   // Starts the Loop, Then Joins It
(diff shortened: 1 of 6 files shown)


ShaneRozenLevy:
I'm all for proper scope, but we shouldn't be including the example code in anything. Does not seem necessary/conventional.

ethanmusser:
This mostly just keeps the namespace clean.  Also keeps the Doxygen-generated documentation organized.

ShaneRozenLevy:
Do you have a link to the doxygen documentation that I can look at?

ethanmusser:
Yep!  Check out the zip file in #33.  This has all the auto-gen HTML and LaTeX.  

Online docs are coming in #34 if we can figure out the deployment authentication.

jdcaporale:
I also thought it was a little weird, but it made some sense for docs and if other labs/projects want to use examples with similar names.

jdcaporale (commented):
not a huge fan of where  `using kodlab::examples::(...) `  is in each of these, but the only other option I can think of is above `int main(...)` and I'm not sure that is better.

ethanmusser:
> not a huge fan of where `using kodlab::examples::(...) ` is in each of these, but the only other option I can think of is above `int main(...)` and I'm not sure that is better.

I don't love it placement-wise, but I wanted to make clear what call it applied to.  Alternately, we can just drop the `using` entirely.

jdcaporale:
> I don't love it placement-wise, but I wanted to make clear what call it applied to. Alternately, we can just drop the `using` entirely.

Probably cleaner honestly.

ethanmusser:
> > I don't love it placement-wise, but I wanted to make clear what call it applied to. Alternately, we can just drop the `using` entirely.
> 
> Probably cleaner honestly.

Changed in 043b9d34a16404ef20649e4cb2fe4a9247a5544f.

ShaneRozenLevy (commented):
I'm neutral on these changes. Doesn't seem necessary, but somewhat beneficial albeit non standard. 

Implementation is sufficient.

ShaneRozenLevy (approved):
LGTM

jdcaporale (approved):

ethanmusser:
I realized that all the example headers would need proper scoping as well.  Not sure how best to organize this.

PR #47, opened 2022-08-06
Lcm multiple subscriber

ethanmusser:
Refactors the `LcmSubscriber` object to permit multiple subscriptions and implements.

- Provides `LcmMessageHandler` object to be derived from or composed into objects requiring inputs.
- Refactors the `LcmSubscriber` object to permit multiple subscriptions via a single `lcm::LCM` object.
- Updates the `MjbotsControlLoop` to compose the new `LcmSubscriber` object as a shared pointer (to be passed around to other objects requiring inputs in the future).
- Updates `robot_example.cpp` and `l_2DoF_example.cpp` to use new input structure.

Note: This multiple-input subscriber will permit behavior-specific LCM inputs in #39.


include/kodlab_mjbots_sdk/lcm_subscriber.h
```diff
@@ -1,85 +1,149 @@
...
-void LcmSubscriber<msg_type>::Run() {
-  lcm_.subscribe(channel_name_, &LcmSubscriber::HandleMsg, this);
-  mutex_.unlock(); // Ensures mutex is unlocked
-  std::cout << "Subscribing to " << channel_name_ << std::endl;
-  while (!CTRL_C_DETECTED) {
-    lcm_.handleTimeout(1000);
+  /**
+   * @brief Map of channel names to corresponding `lcm::Subscription` objects
+   */
+  std::map<std::string, lcm::Subscription *> subs_;
```
jdcaporale, line 133: Presumably we will use this map to unsubscribe by name? Do the subscriptions do anything else?

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -147,7 +148,8 @@ class MjbotsControlLoop : public AbstractRealtimeObject {
   std::string logging_channel_name_;      /// Channel name to publish logs to, leave empty if not publishing
   lcm::LCM lcm_;                          /// LCM object
   LogClass log_data_;                     /// object containing log data
-  LcmSubscriber<InputClass> lcm_sub_;     /// LCM subscriber object
+  LcmSubscriber lcm_sub_;                 /// LCM subscriber object
+  LcmMessageHandler<InputClass> input_sub_;    /// LCM input subscription
```
jdcaporale, line 152: I much prefer this breakout

Still weird we dont have a similar pattern for the log, even though it makes sense mechanically.
This is good. 
Was worried naming here was a little confusing since both of these are kodlab classes, but noticed they are in kodlab namespace, so should be ok.

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -177,7 +181,10 @@ MjbotsControlLoop<log_type, input_type, robot_type>::MjbotsControlLoop(std::vect
 template<class log_type, class input_type, class robot_type>
 MjbotsControlLoop<log_type, input_type, robot_type>::MjbotsControlLoop(std::shared_ptr<robot_type>robot_in, const ControlLoopOptions &options)
   : AbstractRealtimeObject(options.realtime_params.main_rtp, options.realtime_params.can_cpu),
-    lcm_sub_(options.realtime_params.lcm_rtp, options.realtime_params.lcm_cpu, options.input_channel_name) {
+    lcm_sub_(options.realtime_params.lcm_rtp, options.realtime_params.lcm_cpu)
```
jdcaporale, line 184: How many lines before we write a SetupLCMInput()? ;P

examples/robot_example.cpp
```diff
@@ -70,6 +70,7 @@ int main(int argc, char **argv)
     // Define robot options
     kodlab::mjbots::ControlLoopOptions options;
     options.log_channel_name = "motor_data";
+    options.input_channel_name = "mode_input";
```
jdcaporale, line 73: Did we never fix this oversight? or this a default?

include/kodlab_mjbots_sdk/lcm_subscriber.h
```diff
@@ -1,85 +1,149 @@
...
+    subs_.emplace(channel_name, sub);
+    return sub;
+  }
+
+  /**
+   * @brief Accessor for `lcm::Subscription` objects
+   * @param channel_name channel name
+   * @return `lcm::Subscription` object on corresponding channel
+   */
+  [[nodiscard]] const lcm::Subscription *get_subscription(const std::string &channel_name) const
```
jdcaporale, line 118: nodiscard is growing on me, even though its ugly haha

jdcaporale (commented):
LGTM and the different message handling is better than other options.

Couple questions.
Should LcmSubscriber and LcmMessageHandler be broken up into two headers and should either have cpp.

include/kodlab_mjbots_sdk/lcm_subscriber.h
```diff
@@ -1,85 +1,148 @@
...
+   */
+  std::mutex mutex_;
+
+  /**
+   * @brief Constructs an `LcmMessageHandler` and ensures the mutex is unlocked.
    */
-  void HandleMsg(const lcm::ReceiveBuffer *rbuf,
-                 const std::string &chan,
-                 const MessageClass *msg);
+  LcmMessageHandler()
```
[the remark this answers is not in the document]
ethanmusser: Yeah channel name comes in once the subscription itself is created, the handler is strictly the callback and some critical members.

include/kodlab_mjbots_sdk/lcm_subscriber.h
ethanmusser: That was the intention, as well as accessing for customizing the individual `lcm::Subscription` objects.  These can be used to adjust message queue size if desired, as well as unsubscribe.

include/kodlab_mjbots_sdk/mjbots_control_loop.h
ethanmusser: Agreed, I couldn't come to terms with `LcmSubscriber` holding bot the LCM object and the input subscription.

examples/robot_example.cpp
ethanmusser: Haha nope never fixed it, and I run into it every time I try running new input tests!

include/kodlab_mjbots_sdk/lcm_subscriber.h
ethanmusser: Same here, ugly but it does some good!

ethanmusser:
Separating `LcmMessageHandler` and `LcmSubscriber` into separate functions is probably best for code organization.  I will do this once we get @ShaneRozenLevy's feedback on this PR.

include/kodlab_mjbots_sdk/mjbots_control_loop.h
jdcaporale, line 152: `LcmMessageHandler` needs an include technically

include/kodlab_mjbots_sdk/lcm_subscriber.h
```diff
@@ -1,85 +1,102 @@
...
- * @brief An template class for an lcm subscriber. Subscribes to just one msg of type msg_type.
- * @tparam MessageClass the lcm msg type
+/**
+ * @brief LCM subscriber capable of subscribing to multiple channels
+ * @todo Implement subscription removal (i.e., unsubscribing)
  */
-template<class MessageClass>
-class LcmSubscriber : public AbstractRealtimeObject {
- public:
+class LcmSubscriber : public AbstractRealtimeObject
```
ShaneRozenLevy, line 32: since this class is no longer templated, add a .cpp and keep the implementation in the .cpp.

include/kodlab_mjbots_sdk/lcm_subscriber.h
```diff
@@ -1,85 +1,102 @@
...
+#include <map>
+#include <mutex>
 #include <string>
 #include "lcm/lcm-cpp.hpp"
 #include "real_time_tools/thread.hpp"
 #include "kodlab_mjbots_sdk/abstract_realtime_object.h"
+#include "kodlab_mjbots_sdk/lcm_message_handler.h"
+
+namespace kodlab
+{
```
ShaneRozenLevy, line 26: Fix the brackets.

include/kodlab_mjbots_sdk/lcm_subscriber.h
ShaneRozenLevy, line 32: I think its important that we only ever run one of these to prevent concurrency issues.

include/kodlab_mjbots_sdk/lcm_subscriber.h
```diff
@@ -1,85 +1,102 @@
...
+   * @return `lcm::Subscription` object on corresponding channel
    */
-  void Run() override;
+  [[nodiscard]] const lcm::Subscription *get_subscription(const std::string &channel_name) const
+  {
+    return subs_.at(channel_name);
+  }
 
-  std::string channel_name_;
+protected:
```
ShaneRozenLevy, line 76: why protected? Are we expecting to have this class be inherited?

ShaneRozenLevy (changes requested):
2 blockers:

1. Run a test with multiple input lcm data types and makes sure everything works
2. fix the header/source separation and the brackets.

include/kodlab_mjbots_sdk/lcm_subscriber.h
jdcaporale: So right now we only want one on a single thread, partly because of how we are running the handleTimeout(1000) right?

Should we actually include a singleton design pattern to enforce this? something like the threadsafe version found here?
https://stackoverflow.com/questions/1008019/c-singleton-design-pattern

include/kodlab_mjbots_sdk/lcm_subscriber.h
jdcaporale: Philosophically, I feel like I would default to protected unless there is some explicitly reason to protect the interior member, so that it could be derived if someone wanted to. Is this bad practice?

include/kodlab_mjbots_sdk/lcm_subscriber.h
ShaneRozenLevy: no. no non ono no no no. Singletons are horrible. I don't know a single good c++ programmer who likes singeltons.

include/kodlab_mjbots_sdk/lcm_subscriber.h
ShaneRozenLevy: Protected vs private comes into play when you use inheritance with private being the standard. Having it be protected suggests to the reader that this class is expected to be override.

include/kodlab_mjbots_sdk/lcm_subscriber.h
jdcaporale: hahahah 
Why? I have used them before (not in C++ i guess) . 
If the intent is only one should exist, why not enforce that?

jdcaporale: More specifically are they are annoying to implement or dangerous? Or one because the other?

include/kodlab_mjbots_sdk/lcm_subscriber.h
jdcaporale: Yea I guess the point is intent vs capability, with a leaning toward intent. We don't want us or the user to derive so it should be private and it can change if we decide to do something else in the future. Same reason not every destructor is virtual by default.

include/kodlab_mjbots_sdk/lcm_subscriber.h
ShaneRozenLevy: I don't have a great understanding of the hard reasons not to, but I know at least at BDI they are a constant pain to work around and maintain.

jdcaporale: Reading more about it, it seems that the Singleton Pattern implies some level of globality too, which is not what we want. 
I didn't realize the full implications of the pattern in C++. 
I don't think it makes sense in this use case. 

The comments should make it clear though that currently it would be bad to have multiple subscribers on a thread, or a subscriber and anything else really.

include/kodlab_mjbots_sdk/lcm_subscriber.h
ethanmusser: @jdcaporale has it covered, I left it protected in case the user took it upon themselves to make a derived class, but there was no intention of making the user do this.  Will change to private.

include/kodlab_mjbots_sdk/lcm_subscriber.h
ethanmusser: I will update the comments to indicate that concurrency issues may arise if multiple `LcmSubscriber` objects exist simultaneously.

include/kodlab_mjbots_sdk/mjbots_control_loop.h
ethanmusser: Resolved with a3d7d629027fc762a059e3c3f2d441cbf60a1c7e.

include/kodlab_mjbots_sdk/lcm_subscriber.h
ethanmusser: Split in a0b1d74091ed316d4b7e7abb61070ece3161ebc9.

include/kodlab_mjbots_sdk/lcm_subscriber.h
ethanmusser: Fixed in a3d7d629027fc762a059e3c3f2d441cbf60a1c7e

include/kodlab_mjbots_sdk/mjbots_control_loop.h
ethanmusser: Perhaps make an issue?

include/kodlab_mjbots_sdk/lcm_subscriber.h
ethanmusser: Warning provided in 2bfd50c4afc4e757c1dab2850b8696d291b398ef.

include/kodlab_mjbots_sdk/lcm_subscriber.h
ethanmusser: Made private in 2130f57a47c8a4f99b5e132f14dc3884a8383f17.

ethanmusser:
All formatting and header/source convention issues have been resolved.  Will run a proper multi-input test as soon as I get into lab.

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -310,14 +318,14 @@ void MjbotsControlLoop<log_type, input_type, robot_type>::SafeProcessInput() {
...
         ProcessInput();
       }
       // Set new message to false and unlock mutex
-      lcm_sub_.new_message_ = false;
-      lcm_sub_.mutex_.unlock();
+      input_sub_.new_message = false;
+      input_sub_.mutex_.unlock();
     }
   }
 }
```
jdcaporale, line 331: Not an issue that needs to be changed in this PR but something to think about. 
We are handling safety of the input processing here and are using data as a public variable in the input_sub_ class. 
We will likely need a safe process input for every message or at least within any class that has an input message (behavior). Perhaps we could find a structure that makes data private and performs the mutex internally before returning a value copy or a ready flag, 

The solution doesnt have to match for behaviors and control loop, but it would be nice it was possible.

include/kodlab_mjbots_sdk/mjbots_control_loop.h
jdcaporale: I think it fine for now

include/kodlab_mjbots_sdk/lcm_message_handler.h
```diff
@@ -0,0 +1,74 @@
...
+struct LcmMessageHandler {
+  /**
+   * @brief Flag indicating that a new message is available.
+   */
+  bool new_message = false;
+
+  /**
+   * @brief Data for the most-recently received message.
+   */
+  Message data;
```
jdcaporale, line 40: data should probably be a private variable and protected by mutexes for reading. Currently this happens externally via "SafeProcessInput" so that the functions dont take any extra parameters. Mentioned in https://github.com/KodlabPenn/kodlab_mjbots_sdk/pull/47#r939812149

ethanmusser:
A few changes have been made:

- The `LcmSubscriber` object in `MjbotsControlLoop` is now a shared pointer.  This will be handed around to behaviors which will add additional subscribers as needed.
- The `LcmMessageHandler` has been turned into a class with proper access protections, particularly in regards to `data_`, which may be mutex locked at access time.


The complete PR has been tested and is functioning with several inputs.

examples/robot_example.cpp
```diff
@@ -45,9 +45,9 @@ class SimpleRobotControlLoop : public kodlab::mjbots::MjbotsControlLoop<ManyMoto
         }
     }
 
-    void ProcessInput() override
+    void ProcessInput(const ModeInput &input_data) override
```
jdcaporale, line 48: I like this cause it forces the user to be explicit and reminded about their type and also has better memory control.  win win in my book, even if it isn't quite as easy to use templatewise.

include/kodlab_mjbots_sdk/lcm_subscriber.h
```diff
@@ -1,85 +1,121 @@
...
   cpu_ = cpu;
   realtime_priority_ = realtime_priority;
-  if (!channel_name_.empty())
-    Start();
+  Start();  // FIXME: Call outside constructor
+}
+
+void LcmSubscriber::RemoveSubscription(const std::string& channel_name) {
+  lcm_.unsubscribe(subs_[channel_name]);
+  subs_.erase(subs_.find(channel_name));
```
jdcaporale, line 107: This works but I think that erase also takes just the key itself ie
```suggestion
  subs_.erase(channel_name);
```
which is easier to read

jdcaporale: Reading more about it it seems slightly more complicated, but since we are using the [] operator immediately above we have inherently assumed that channel name will be in subs_ map. 

Likely though we should check it is in the map and then interact with it? https://stackoverflow.com/a/33239463

include/kodlab_mjbots_sdk/lcm_subscriber.h
```diff
@@ -1,85 +1,121 @@
...
+}
+
+[[nodiscard]] const lcm::Subscription *LcmSubscriber::get_subscription(const std::string &channel_name) const {
+  return subs_.at(channel_name);
```
jdcaporale, line 111: Throws exception if channel name is not available, not sure how well our system handles non-CtrlC exceptions

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -147,7 +150,8 @@ class MjbotsControlLoop : public AbstractRealtimeObject {
   std::string logging_channel_name_;      /// Channel name to publish logs to, leave empty if not publishing
   lcm::LCM lcm_;                          /// LCM object
   LogClass log_data_;                     /// object containing log data
-  LcmSubscriber<InputClass> lcm_sub_;     /// LCM subscriber object
+  std::shared_ptr<LcmSubscriber> lcm_sub_;     /// LCM subscriber object
+  LcmMessageHandler<InputClass> input_sub_;    /// LCM input subscription
```
jdcaporale, line 154: Naming here is awkward, but fine.

include/kodlab_mjbots_sdk/lcm_message_handler.h
```diff
@@ -0,0 +1,109 @@
...
+    data_ = *msg;
+    new_message_ = true;
+    mutex_.unlock();
+  }
+
+  /**
+   * @brief Retrieve new data, if available.
+   * @details Retrieve a `std::optional` object containing new data if the
+   * internal `new_message_` flag evaluates to `true`, or `std::nullopt_t`
+   * otherwise. If new data is retireved, the new message flag is reset to
```
jdcaporale, line 69: Typo

include/kodlab_mjbots_sdk/lcm_message_handler.h
```diff
@@ -0,0 +1,109 @@
...
+   * `false`.
+   * @return std::optional containing message data if new message available,
+   * std::nullopt_t otherwise
+   */
+  std::optional<Message> GetNewData() {
```
jdcaporale, line 74: Possible rename
```suggestion
  std::optional<Message> GetDataIfNew() {
```
Would allow for similarly named , `GetData()`,

jdcaporale: Beginning of possible implementation. Not necessary for this PR since we never use latent data in-between input updates.

 ```
template<Message>
const Message GetData(){
    std::optional<Message> data_out;
    data_out = GetDataIfNew()
    if(data_out){
        return data_out.value();
    }
    else{
        return data_cached; // Needs another value to avoid mutex with the assumption 
                                        // that data cached is entirely in a single thread and needs no mutex.
    }

 ```

include/kodlab_mjbots_sdk/lcm_message_handler.h
```diff
@@ -0,0 +1,109 @@
...
+   * @brief Virtual destructor for clean inherited class destruction
+   */
+  virtual ~LcmMessageHandler() = default;
+
+  /**
+   * @brief Callback function for when an LCM message is received.
+   * @details This function serves as a callback for incoming LCM messages on
+   * `channel`.  It copies the decoded message data to `data`, and sets
+   * `new_message` to `true`.
+   * @param rbuf LCM recieve buffer containing raw bytes and timestamp
```
jdcaporale, line 52: typo

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -125,10 +126,12 @@ class MjbotsControlLoop : public AbstractRealtimeObject {
   void SafeProcessInput();
 
   /*!
-   * @brief virtual class to be implemented when logging. Process data in m_lcm_sub.m_data; This function is threadsafe
-   * and won't run if the LCM thread holds the mutex
+   * @brief virtual class to be implemented when logging. Process data in
```
jdcaporale, line 129: function not class

jdcaporale: function member or method I guess

jdcaporale (approved):
Some changes and comments. Goal posts are moving cause we are able to make code that is so much better now. I still marked as approve, cause most of these are suggestions more than actual issues. 

Few typos and couple questions about map, but I don't think I need to look at it again after you fix these things

include/kodlab_mjbots_sdk/mjbots_control_loop.h
ethanmusser: Handled in ba8dcf1ad7d58d23fd9b58d79243d8c08754743c

include/kodlab_mjbots_sdk/lcm_message_handler.h
ethanmusser: Resolved in ba8dcf1ad7d58d23fd9b58d79243d8c08754743c

examples/robot_example.cpp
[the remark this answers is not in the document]
ethanmusser: Areed on both fronts.  All around clearer.

include/kodlab_mjbots_sdk/lcm_subscriber.h
```diff
@@ -1,85 +1,121 @@
...
-/*!
- * @brief An template class for an lcm subscriber. Subscribes to just one msg of type msg_type.
- * @tparam MessageClass the lcm msg type
+
+/**
+ * @brief LCM subscriber capable of subscribing to multiple channels
+ * @warning Running multiple `LcmSubscriber` objects on a single CPU can result
+ * in concurrency issues.  The authors do not endorse this usage, do so at your
+ * own risk.
+ * @todo Move implementation to source file once `CTRL_C_DETECTED` global
```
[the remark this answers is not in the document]
ethanmusser: We should propagate to other relevant classes, if any (eyeing `MjbotsControlLoop`).

include/kodlab_mjbots_sdk/lcm_subscriber.h
```diff
@@ -1,85 +1,121 @@
...
+
+  /**
+   * @brief LCM object
+   */
   lcm::LCM lcm_;
+
+  /**
+   * @brief Map of channel names to corresponding `lcm::Subscription` objects
+   */
+  std::map<std::string, lcm::Subscription *> subs_;
```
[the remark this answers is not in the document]
ethanmusser: Valid Doxygen block comments are Javadoc style (`/** ... */`), Qt style (`/*! ... */`), or triple-slash blocks (`///` on every line).  I've been sticking with Javadoc style throughout my contribs, but maybe we should standardize.

The alternative are inline, triple-slash comments (i.e., `///` on the same line as the member), but these only permit a partial-line's-worth or text and no room for additional flags.  I think the choice between inline and blocks is just circumstantial.

See [Doxygen reference](https://doxygen.nl/manual/docblocks.html) for supported styles.

include/kodlab_mjbots_sdk/mjbots_control_loop.h
ethanmusser: Amended in e73960cbc50c2c7567521410a70b84a7d077c952

include/kodlab_mjbots_sdk/lcm_message_handler.h
ethanmusser: Resolved in 20b0337a22b139e39944e7f055d572ed1ff98f62

include/kodlab_mjbots_sdk/lcm_message_handler.h
ethanmusser: Naming changed in 20b0337a22b139e39944e7f055d572ed1ff98f62.  We still have some mutex issues to deal with when caching data, let's save this for the next PR.  Maybe make an issue with this snippet.

include/kodlab_mjbots_sdk/lcm_message_handler.h
ethanmusser: Resolved in 20b0337a22b139e39944e7f055d572ed1ff98f62

include/kodlab_mjbots_sdk/mjbots_control_loop.h
ethanmusser: I tried to keep with convention, may be worth changing in the future.  For now, I think this is sufficient.

include/kodlab_mjbots_sdk/lcm_subscriber.h
ethanmusser: Handled in 5accfd78db2fcd592f411818e7780800fed596e8.  now returns `nullptr` if channel not in `std::map`

include/kodlab_mjbots_sdk/lcm_subscriber.h
ethanmusser: Handled invalid channel entries in 5accfd78db2fcd592f411818e7780800fed596e8.

include/kodlab_mjbots_sdk/lcm_subscriber.h
```diff
@@ -1,85 +1,121 @@
...
+////////////////////////////////////////////////////////////////////////////////
+// Implementation                                                             //
+////////////////////////////////////////////////////////////////////////////////
+
+LcmSubscriber::LcmSubscriber(int realtime_priority, int cpu) {
   cpu_ = cpu;
   realtime_priority_ = realtime_priority;
-  if (!channel_name_.empty())
-    Start();
+  Start();  // FIXME: Call outside constructor
```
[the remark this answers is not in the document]
ethanmusser: Start moved in 978857ac3371bcf9d2096e39b157cd6dd1f4d14b.  `cpu_` and `realtime_priority_` are both PODs so this should be fine as is.

PR #50, opened 2022-08-08
Lcm publisher

ethanmusser:
Provide `LcmPublisher` object to handle storage and publishing of LCM messages.  This object makes it easy to publish to multiple LCM channels via one or multiple `lcm::LCM` objects, and handle storage of this information.

- Provides `LcmPublisher` object
- Replaces usage of `lcm::LCM` object in `MjbotsControlLoop` with `LcmPublisher` object.
- Updates `MjbotsControlLoop`-derived control loop to use new logging structure.

### DIFF
--- a/examples/imu_example.cpp
+++ b/examples/imu_example.cpp
@@ -64,30 +64,30 @@ private:
   }
 
   /**
-   * @brief Prepares log
+   * @brief Populates log message with data from current control loop cycle
    */
   void PrepareLog() override
   {
     // IMUData Quaternion
-    log_data_.attitude_quaternion[0] = imu_->get_quat().x();
-    log_data_.attitude_quaternion[1] = imu_->get_quat().y();
-    log_data_.attitude_quaternion[2] = imu_->get_quat().z();
-    log_data_.attitude_quaternion[3] = imu_->get_quat().w();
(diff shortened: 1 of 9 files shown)


ShaneRozenLevy (approved):
LGTM.

jdcaporale (approved):
LGTM, barring a minor discussion (not action) point.
I'm conflicted on the use of log_data_ and log_pub_ on the same level as members of the same class, since it fails to match input pattern.
I don't feel that strongly about it .

ethanmusser:
Fixed a seg fault issue, all working with multiple publishers.  

@jdcaporale, I see what you're saying, but I can't think of a better way to delegate editing of the message data to the user with specific access to message elements without making them populate the entire message and copy it over to `LcmPublisher`.  I think this is preferrable, but perhaps the other way is safer.  Either way, both message data and publisher will have to live on the same level in `MjbotsControlLoop`.

ethanmusser:
Fixed a seg fault issue, all working with multiple publishers.  

@jdcaporale, I see what you're saying, but I can't think of a better way to delegate editing of the message data to the user with specific access to message elements without making them populate the entire message and copy it over to `LcmPublisher`.  I think this is preferrable, but perhaps the other way is safer.  Either way, both message data and publisher will have to live on the same level in `MjbotsControlLoop`.

PR #52, opened 2022-08-11
Add channel selection to mode_input_cli.py

ethanmusser:
Simple change permitting channel selection when running the sample `mode_input_cli.py` script for sending mode inputs.  The script should now be called by
```bash
python3 mode_input_cli.py <channel_name>
```

### DIFF
--- a/python_scripts/mode_input_cli.py
+++ b/python_scripts/mode_input_cli.py
@@ -22,6 +22,7 @@ Copyright:
 import lcm
 from lcm_types.ModeInput import ModeInput
 import warnings
+import argparse
 
 # Ignore deprecation warnings from ``lcm`` dependencies
 warnings.filterwarnings("ignore", category=DeprecationWarning)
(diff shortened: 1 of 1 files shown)


python_scripts/mode_input_cli.py
```diff
@@ -22,17 +22,30 @@
...
 OFF = 0  # OFF behavior integer
 KILL = -1  # KILL behavior integer
-LCM_CHANNEL = "mode_input"  # must match channel name in control loop options
+DEFAULT_LCM_CHANNEL = "mode_input"
 
 
-def publish_mode(mode, print_msg="", lcm_ch=LCM_CHANNEL):
+def parse_channel():
+    """
+    """
```
jdcaporale, line 38: This would need a comment

python_scripts/mode_input_cli.py
```diff
@@ -22,17 +22,30 @@
...
+    # Parse input arguments
+    parser = argparse.ArgumentParser(
+        description='publish mode input messages via LCM.')
+    parser.add_argument('channel', metavar='channel', type=str, nargs='+',
+                        help='LCM channel name to publish on')
+    args = parser.parse_args()
+    return args.channel[0]
```
jdcaporale, line 45: To allow people to use this to its fullest I would likely set this up to return the args and then be used as args in the main()
This would allow us or others to add arguments more easily, for example multi channel, or rate, or something like that.

jdcaporale: A little more work than mode_input needs but easier to make into whatever you want

jdcaporale (approved):
LGTM but missing a comment and then an optional change that I think would make things cleaner in the future.

python_scripts/mode_input_cli.py
ethanmusser: Fair enough, fixed in 898235f93615e204682187e0ba2711b6676ef07e

python_scripts/mode_input_cli.py
ethanmusser: Added in 90886ddd9ff97e4bd0db4156b6bb8767f09651c6

PR #58, opened 2022-08-19
Fix BehaviorManager segfault when new behavior initialization fails

ethanmusser:
Fix a segmentation fault in `BehaviorManager` when `ConstructBehavior` is run and fails to initialize a behavior.  Tested on the motor test setup.


jdcaporale (approved):
LGTM, for hotfix.
Begs the question: Should ConstructBehavior also Init the Behavior or just construct and output it? And then the safe Init happens elsewhere or after that call.

PR #59, opened 2022-08-20
Implement dry run capability

ethanmusser:
Provides a dry run boolean flag in the `MjbotsControlLoopOptions` struct that runs the control loop and prints the commanded torques but sends zero-torques to the motors.

### DIFF
--- a/src/mjbots_hardware_interface.cpp
+++ b/src/mjbots_hardware_interface.cpp
@@ -9,6 +9,8 @@
 #include <iostream>
 #include <algorithm>
 
+#include "kodlab_mjbots_sdk/log.h"
+
 namespace kodlab::mjbots {
 void MjbotsHardwareInterface::InitializeCommand() {
   for (const auto &joint : joints) {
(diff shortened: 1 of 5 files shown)


jdcaporale (approved):
LGTM, except an errant whitespace, but technically it is better than before

src/mjbots_hardware_interface.cpp
```diff
@@ -118,9 +120,18 @@ void MjbotsHardwareInterface::ProcessReply() {
...
+
+  if (!dry_run_) {
+    for (int servo = 0; servo < num_joints_; servo++) {// TODO Move to a seperate update method (allow non-ff torque commands)?
+      commands_[servo].position.feedforward_torque = joints[servo]->get_servo_torque();
+    }
+  } else {
+    std::fprintf(stdout, "Torques: ");
+    for (int servo = 0; servo < num_joints_; servo++) {
+      commands_[servo].position.feedforward_torque = 0.0;
+      std::fprintf(stdout, "% 8.2f, ", joints[servo]->get_servo_torque());
```
ShaneRozenLevy, line 132: Why are we using print and not the logging macros?

ShaneRozenLevy (approved):
Seems fine.

src/mjbots_hardware_interface.cpp
jdcaporale: I think because it is a vector

ethanmusser: In this case laziness.  I wanted to print a full each element of the `std::vector` on a single line, and we do not currently have a macro for printing vectors of values.  Might be a good addition though.

jdcaporale: Also, I think that I am going to add a print_torques bool, because as expected, printing is a little aggressive.

jdcaporale: @ShaneRozenLevy Thoughts or revert?

ShaneRozenLevy: Probably fine. I don't have super strong feelings about this.

ethanmusser: Fixed in d600e77fe0a5fce1e9499f2b68c09096d39340f3

PR #61, opened 2022-08-22
Add troubleshooting section to Readme

chandravaran:
On 18.04 got errors for pthread, please look at the troubleshooting section to fix these errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,10 +287,29 @@ Normal build is
 
     cmake .. -DCMAKE_TOOLCHAIN_FILE=~/mjbots/kodlab_mjbots_sdk/cmake/pi.cmake
 
+
+
 # Running Code
 To Run code, first scp the binary onto the pi, and then Run it as sudo using:
 
     sudo ./code_binary
+
+# Troubleshooting 
+If you encounter the below a pthread error during Cmake keep the below commands in the main repo directory CMakeLists.txt after the endif()
+
+    set(CMAKE_THREAD_LIBS_INIT "-lpthread")
+    set(CMAKE_HAVE_THREADS_LIBRARY 1)
+    set(CMAKE_USE_WIN32_THREADS_INIT 0)
+    set(CMAKE_USE_PTHREADS_INIT 1)
+    set(THREADS_PREFER_PTHREAD_FLAG ON)
+
+If you run into another pthread error while running the "make" command then you can run the following command in a terminal to fix it
+   
+    cd <path to rpi folder's rootfs>/usr/lib/arm-linux-gnueabihf; ln -s ../../../lib/arm-linux-gnueabihf/libpthread.so.0 libpthread.so
+
+The reason you could have got an error while making is because there was a broken link on the rpi folders you rsynced or zip file you got.
+
+Fix can be found here: https://github.com/Azure/azure-iot-sdk-c/issues/1093
 
 # Citation
 To cite this repo please use the information in the CITATION.cff file


README.md
```diff
@@ -287,11 +287,30 @@ Normal build is
...
+
 # Running Code
 To Run code, first scp the binary onto the pi, and then Run it as sudo using:
 
     sudo ./code_binary
 
+# Troubleshooting 
+If you encounter the below a pthread error during Cmake keep the below commands in the main repo directory CMakeLists.txt after the endif()
+
+    set(CMAKE_THREAD_LIBS_INIT "-lpthread")
```
ShaneRozenLevy, line 300: Why are these changes not made in the `CMakeList.txt`? Does it break it for other people?

README.md
```diff
@@ -287,11 +287,30 @@ Normal build is
...
+    set(CMAKE_HAVE_THREADS_LIBRARY 1)
+    set(CMAKE_USE_WIN32_THREADS_INIT 0)
+    set(CMAKE_USE_PTHREADS_INIT 1)
+    set(THREADS_PREFER_PTHREAD_FLAG ON)
+
+If you run into another pthread error while running the "make" command then you can run the following command in a terminal to fix it
```
ShaneRozenLevy, line 306: This isn't clear. Its probably better to suggest redoing the rsync. Also be explicit about what sort of errors you saw.

README.md
```diff
@@ -287,11 +287,30 @@ Normal build is
...
     cmake .. -DCMAKE_TOOLCHAIN_FILE=~/mjbots/kodlab_mjbots_sdk/cmake/pi.cmake
 
+
+
 # Running Code
 To Run code, first scp the binary onto the pi, and then Run it as sudo using:
 
     sudo ./code_binary
 
+# Troubleshooting
```
ShaneRozenLevy, line 297: Probably this should not be at this level since its troubleshooting the setup.

README.md
```diff
@@ -287,11 +287,30 @@ Normal build is
...
+
+    set(CMAKE_THREAD_LIBS_INIT "-lpthread")
+    set(CMAKE_HAVE_THREADS_LIBRARY 1)
+    set(CMAKE_USE_WIN32_THREADS_INIT 0)
+    set(CMAKE_USE_PTHREADS_INIT 1)
+    set(THREADS_PREFER_PTHREAD_FLAG ON)
+
+If you run into another pthread error while running the "make" command then you can run the following command in a terminal to fix it
+   
+    cd <path to rpi folder's rootfs>/usr/lib/arm-linux-gnueabihf; ln -s ../../../lib/arm-linux-gnueabihf/libpthread.so.0 libpthread.so
```
ShaneRozenLevy, line 308: If you want to include this fix probably better to just link to the thread rather than including a copy of the fix, but lacking context.

ShaneRozenLevy (commented):
I think a troubleshooting section is reasonable, but its important to make it useful and readable and not just your notes from when you setup the code.

README.md
chandravaran: These changes don't break it, but @ethanmusser said that as we are looking to support the sdk for 20.04 and these occurred on 18.04 they are not required in the CMakeList.txt.

README.md
chandravaran: As the rsync was slow, I had received the rpi files from @zacgong, which he got from rpi. That could be a reason for the broken syslink. This needs to be checked if it occurs on either of the pi's with just plain rsync and not with copying of the files.

README.md
chandravaran: Should I shift it one section above to, right after the build?

README.md
chandravaran: The thread is kind of huge and had a few solutions so I kept the one solution that resolved it for me. However, I have given the link for the thread below that line.

README.md
ShaneRozenLevy: If it doesn't break 20.04, then you should add it to the `CmakeList.txt`. It'll be a pain for you to develop with that unmerged change.

README.md
ShaneRozenLevy: The number of `#` denotes the level, so move it to `##`, then put it under in the lines after `motor setup` which is the last thing beneath `setup`

chandravaran: Okay will do that

PR #64, opened 2022-08-26
Limb class

zacgong:
added a LimbBase class inbetween JointBase and RobotBase, abstract class with get/set position velocity and torques for a limb of joints, and virtual functions for kinematics.  Also added very basic limb functionality to RobotBase.  

TO BE DONE: Add a SimpleLimb example, then a 3DOFLimb example with implemented kinematics


ShaneRozenLevy:
Generally when you close a pr, that you have tagged other people on, its good etiquette to leave a brief comment as to why, i.e. closed because I did not mean to open it.

Issue #51, opened 2022-08-09
CTRL_C_DETECTED Scope


ethanmusser:
`kodlab::CTRL_C_DETECTED` is defined as a static, gobal variable in `common_header.h`.  As such, it is restircted to the scope of the file in wihch it is defined (i.e., it is not available externally to other translation units), and cannot be used in `cpp` files other than that where then main method is defined.  I am sure access to this variable should not be restricted to header files.  

### Expected Behavior

`kodlab::CTRL_C_DETECTED` is accessible globally in all source `cpp` files. 

### Current Behavior

`kodlab::CTRL_C_DETECTED` is only available in the same translation unit as the main method (i.e., not in any other source `cpp` files). 

### Proposed Solutions

- `kodlab::CTRL_C_DETECTED` should be declared as a non-static, `extern` in `common_header.h` and defined in some common header implementation file (as done at the top of [this answer](https://stackoverflow.com/a/1433387)).
- We should provide more clear methods for setting and reading `kodlab::CTRL_C_DETECTED`.

jdcaporale:
Fixed by #87

PR #65, opened 2022-08-27
Inline LcmSubscriber implementation

ethanmusser:
Currently, the `LcmSubscriber` implementation is in the `lcm_subscriber.h` header file where the class is defined.  If this file is included in multiple translation units, a linker error occurs due to multiple definitions of the member functions.  This will be fixed once the TODO in `lcm_subscriber.h` requesting that the implementation be moved to a `cpp` file is addressed.  However, this cannot happen until #51 is resolved.  This is a temporary fix to prevent linker errors in the meantime.

### DIFF
--- a/include/kodlab_mjbots_sdk/lcm_subscriber.h
+++ b/include/kodlab_mjbots_sdk/lcm_subscriber.h
@@ -106,16 +106,16 @@ class LcmSubscriber : public AbstractRealtimeObject {
 // Implementation                                                             //
 ////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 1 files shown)


ShaneRozenLevy (approved):
LGTM. I don't see how this will be improved by #51 since the `LCMSubscriber` is a templated class with an unknowable array of possible template arguments.

ethanmusser:
@ShaneRozenLevy That is a good point.  The class itself and most of it's definitions can move to a `cpp`, but there is one member that is a function template.  I guess we're stuck with keeping everything inlined in the header since it would be impossible to explicitly instantiate the method template for all LCM message types.

PR #69, opened 2022-09-23
Added basic find joint by name functionality

ShaneRozenLevy:
Since we have joint names when we want them I thought it worthwhile to add some basic functionality for finding a joint by name rather than by index.

### DIFF
--- a/include/kodlab_mjbots_sdk/robot_base.h
+++ b/include/kodlab_mjbots_sdk/robot_base.h
@@ -181,6 +181,6 @@ namespace kodlab
         real_time_tools::Timer run_timer_;                            /// Run timer for robot, started at construction
         SoftStart soft_start_;                                        /// Soft Start object
         int num_joints_ = 0;                                          /// Number of joints
-        std::unordered_map<std::string, int> joint_name_to_index_;    /// Map from joint name to joint index
+        std::unordered_map<std::string, int> joint_name_to_index_;    ///< Map from joint name to joint index
     };
 } // namespace kodlab
--- a/include/kodlab_mjbots_sdk/robot_base.h
+++ b/include/kodlab_mjbots_sdk/robot_base.h
@@ -50,7 +50,10 @@ namespace kodlab
                 joints.push_back(j); // Copy constructed, uses an implicit upcasting
                 positions_.push_back( j->get_position_reference() );
                 velocities_.push_back( j->get_velocity_reference() );
-                torque_cmd_.push_back( j->get_servo_torque_reference() ); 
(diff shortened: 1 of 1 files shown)


include/kodlab_mjbots_sdk/robot_base.h
```diff
@@ -171,5 +181,6 @@ namespace kodlab
         real_time_tools::Timer run_timer_;                            /// Run timer for robot, started at construction
         SoftStart soft_start_;                                        /// Soft Start object
         int num_joints_ = 0;                                          /// Number of joints
+        std::unordered_map<std::string, int> joint_name_to_index_;    /// Map from joint name to joint index
```
ethanmusser, line 184: For the [Doxygen to work properly](https://doxygen.nl/manual/docblocks.html):
```suggestion
        std::unordered_map<std::string, int> joint_name_to_index_;    ///< Map from joint name to joint index
```

include/kodlab_mjbots_sdk/robot_base.h
```diff
@@ -50,7 +50,10 @@ namespace kodlab
                 joints.push_back(j); // Copy constructed, uses an implicit upcasting
                 positions_.push_back( j->get_position_reference() );
                 velocities_.push_back( j->get_velocity_reference() );
-                torque_cmd_.push_back( j->get_servo_torque_reference() ); 
+                torque_cmd_.push_back( j->get_servo_torque_reference() );
+                if(j->get_name() != ""){
+                  joint_name_to_index_.insert({j->get_name(), joints.size()-1});
```
ethanmusser, line 55: Might be worth checking if the name already exists in `joint_name_to_index_` and warning about duplicate naming if it does.

ethanmusser (approved):
Not sure if you want to add a user warning if there are duplicate joint names, but this could be good to avoid accessing an unintended joint. Otherwise looks good.

include/kodlab_mjbots_sdk/robot_base.h
ShaneRozenLevy: Map insert fails due to a duplicate entry

ethanmusser: Fails but does not output any warning.  I meant to add a check similar to `ok` in the [example here](https://en.cppreference.com/w/cpp/container/unordered_map/insert).  Either way, NBD.

ShaneRozenLevy: Yeah. I'm adding a LOG_ERROR in a follow up pr for duplicate names. I'll hold off on merging it in since I should actually test it.

PR #70, opened 2022-09-23
Changes where log plotters look for logs for easier tab completion

ShaneRozenLevy:
I've found it helpful on my own log plotters to have the log file input include the path making it possible to tab complete. Thoughts?


ethanmusser (approved):
Huh, I could have sworn that either @jdcaporale or I had already done that!  Must just be in our Twist stuff.  Good with me!

PR #72, opened 2022-09-28
Throw error on duplicate joint names

ShaneRozenLevy:
Follow up from #69 that throws an error when duplicate joint names.


ethanmusser (approved):
LGTM

PR #73, opened 2022-09-29
Enabled dry run and stop for pd commands

ShaneRozenLevy:
I realized that in #71 I didn't properly hook up the PD commands to the dry run flag and the stop function in RobotBase. This PR fixes that.


ethanmusser (approved):
Good catch, looks good.

Issue #63, opened 2022-08-26
PD/PID controller class


ethanmusser:
Feature request for a proportional-derivative (or perhaps proportional-integral-derivative) controller class.

### Desired Behavior

Users can utilize the PD/PID controller class for applicable use-cases, or implement their own, more advanced controllers through child classes.

### Questions

- I am not sure if this should be a PD or PID controller implementation.  For Kod*Lab applications, we generally only require PD.  Any thoughts?

### Minimal Proposed Implementation

- `Controller`: An abstract class containing generic functions applicable to controllers, including:
   - Update - Compute controller output from the latest data
- `PDController` or `PIDController`: A child class to `Controller` that implements a [discrete-time PD/PID controller](https://en.wikipedia.org/wiki/Low-pass_filter#Discrete-time_realization).

chandravaran:
![Block_diagram](https://user-images.githubusercontent.com/50416143/194768311-f9ab8b7c-1ff6-47a2-b83d-d26ad81d4947.png)

I was thinking these are the functions and variables that could be in abstract class

ethanmusser:
> ![Block_diagram](https://user-images.githubusercontent.com/50416143/194768311-f9ab8b7c-1ff6-47a2-b83d-d26ad81d4947.png)
> 
> 
> 
> I was thinking these are the functions and variables that could be in abstract class

Some thoughts:
- It's not really clear to me what the `Start`, `Run`, and `Stop` methods are intended to do here.
- `Init` only really makes sense as a factory constructor a la [this](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-factory).
- How do you intend to make the signature for the virtual `set_gains` function generic to any kind of controller gains (different controller implementations will have different types and numbers of gains).

chandravaran:
So I was thinking there could be times when we want to shift from one controller scheme to another. So in that case we would require to start, run and stop. Which is why I kept these items. If I am not wrong if we were to template the function we would be able to send n number of parameters. Or we could send a vector and then do a check based on the number of parameters. Also we could use a 2-dimensional array that has the names as well.

ethanmusser:
[...]

PR #74, opened 2022-09-30
Created the inital implementation of the PID controller class

chandravaran:
Introduces an abstract `Pid` class.
This is a new feature. Addresses #63 .

### Change List

- Provides `Pid` object, a class for implementing discrete-time pid control. 

### Upcoming Changes

- Creating an abstract class so various controllers can be derived. 

### Testing

This PR draft has yet to be tested. 

### Questions

How we should proceed in developing the templates for the abstract controller we will have and this class as well.

### DIFF
--- a/include/kodlab_mjbots_sdk/pid.h
+++ b/include/kodlab_mjbots_sdk/pid.h
@@ -1,0 +1,192 @@
+/**
+ * @file pid.h
+ * @author Chandravaran Kunjeti (kunjeti@seas.upenn.edu)
+ * @brief Class implementation of PID
+ * @date 9/30/2022
+ *
+ * @copyright 2022 The Trustees of the University of Pennsylvania. All rights reserved.
+ *
+ */
+
+#pragma once
+
+#ifndef _PID_
+#define _PID_
+
+#include "Eigen/Geometry"
+
+namespace kodlab
+{
+
+  /**
+   * @brief Pid object
+   */
+  class Pid
+  {
+
+  public:
+    /**
+     * @brief Construct a Pid object
+     * @param p_gain Proportional Gain
+     * @param d_gain Derivative Gain
+     * @param i_gain Integral Gain
+     * @param time_step Time difference between 2 updates
+     */
+    Pid(double p_gain, double d_gain, double i_gain, double time_step, double deadband, double saturation)
+        : kp_(0), kd_(0), ki_(0), time_step_(0.01), deadband_(0), saturation_(0)
+    {
+
+      kp_ = p_gain;
+      kd_ = d_gain;
+      ki_ = i_gain;
+      time_step_ = time_step;
+      deadband_ = deadband;
+      saturation_ = saturation;
+    }
+
+    /**
+     * @brief Function to set the PID gains
+     * @param p_gain Proportional Gain
+     * @param d_gain Derivative Gain
+     * @param i_gain Integral Gain
+     */
+    void SetGains(double p_gain, double d_gain, double i_gain)
+    {
+
+      kp_ = p_gain;
+      kd_ = d_gain;
+      ki_ = i_gain;
+    }
+
+    /**
+     * @brief Function to set the deadband
+     * @param deadband Deadband that needs to be set
+     */
+    void SetDeadband(double deadband)
+    {
+
+      deadband_ = deadband;
+    }
+
+    /**
+     * @brief Function to set the Saturation
+     * @param saturation Saturation
+     */
+    void SetSaturation(double saturation)
+    {
+
+      saturation_ = saturation;
+    }
+
+    /**
+     * @brief Function to update the PID output.
+     * @param goal Final goal state for the controller.
+     * @param current Current state of the controller.
+     * @return Returns the Pid output
+     */
+    double Update(double goal, double current)
+    {
+
+      // Calculate the errors
+      // Proportional error
+      error_ = goal - current;
+
+      // Integral error
+      i_error_ += error_ * time_step_;
+
+      // Derivative error
+      derivative_ = (error_ - d_error_) / time_step_;
+
+      // Control output
+      output_ = kp_ * error_ + kd_ * derivative_ + ki_ * i_error_;
+
+      // Saturation checking
+      output_ = (output_ > saturation_) ? saturation_ : output_;
+
+      // Storing the past error
+      d_error_ = derivative_;
+
+      return output_;
+    }
+
+    /**
+     * @brief Function to update the PID output given the respective errors
+     * @param error_p Proportional Error
+     * @param error_d Derivative Error
+     * @param error_i Integral Error
+     * @return Returns the Pid output
+     */
+    double UpdateWithError(double error_p, double error_d, double error_i)
+    {
+
+      // Finding the control output if errors are inputed
+      output_ = kp_ * error_p + kd_ * error_d + ki_ * error_i;
+      output_ = (output_ > saturation_) ? saturation_ : output_;
+      return output_;
+    }
+
+  protected:
+    /**
+     * @brief Time difference between 2 updates
+     */
+    double time_step_;
+
+    /**
+     * @brief Proportional Gain
+     */
+    double kp_;
+
+    /**
+     * @brief Derivative Gain
+     */
+    double kp_;
+
+    /**
+     * @brief Derivative Gain
+     */
+    double kd_;
+
+    /**
+     * @brief Integral Gain
+     */
+    double ki_;
+
+    /**
+     * @brief Stores the previous error for the d term
+     */
+    double d_error_ = 0;
+
+    /**
+     * @brief Current derivative error found
+     */
+    double derivative_;
+
+    /**
+     * @brief Sumes up the error for calculating the I term
+     */
+    double i_error_;
+
+    /**
+     * @brief Error at each step.
+     */
+    double error_;
+
+    /**
+     * @brief Deadband of the system
+     */
+    double deadband_;
+
+    /**
+     * @brief Max control output
+     */
+    double saturation_;
+
+    /**
+     * @brief Output of the function
+     */
+    double output_;
+  };
+
+} // kodlab::mjbots
+
+#endif


include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,192 @@
...
+    Pid(double p_gain, double d_gain, double i_gain, double time_step, double deadband, double saturation)
+        : kp_(0), kd_(0), ki_(0), time_step_(0.01), deadband_(0), saturation_(0)
+    {
+
+      kp_ = p_gain;
+      kd_ = d_gain;
+      ki_ = i_gain;
+      time_step_ = time_step;
+      deadband_ = deadband;
+      saturation_ = saturation;
```
ethanmusser: Why are you using an initialization list (with magic number initialization values) and then re-assigning values to these same data members in the constructor body?  This needlessly writes to these members twice (with the initial value immediately discarded).

[This section of the ISO C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c49-prefer-initialization-to-assignment-in-constructors) indicates that an initialization list is the preferred method here.  But, of course, this should initialize with the passed-in parameter values.

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,192 @@
...
+      kp_ = p_gain;
+      kd_ = d_gain;
+      ki_ = i_gain;
+    }
+
+    /**
+     * @brief Function to set the deadband
+     * @param deadband Deadband that needs to be set
+     */
+    void SetDeadband(double deadband)
```
ethanmusser: Mutators and accessors (getters and setters) should be in snake case per [this section of the style guide](https://drake.mit.edu/styleguide/cppguide.html#Function_Names).

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,192 @@
...
+      derivative_ = (error_ - d_error_) / time_step_;
+
+      // Control output
+      output_ = kp_ * error_ + kd_ * derivative_ + ki_ * i_error_;
+
+      // Saturation checking
+      output_ = (output_ > saturation_) ? saturation_ : output_;
+
+      // Storing the past error
+      d_error_ = derivative_;
```
ethanmusser: Why are all of these computed and saved as class members?  I feel like more information is being retained here than necessary.

ethanmusser: There is duplicate code in these lines of `Update`.  This should instead call `UpdateWitherror`.

ethanmusser: This assignment can happen above, no need for intermediate `derivative_` variable.

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,192 @@
...
+      i_error_ += error_ * time_step_;
+
+      // Derivative error
+      derivative_ = (error_ - d_error_) / time_step_;
+
+      // Control output
+      output_ = kp_ * error_ + kd_ * derivative_ + ki_ * i_error_;
+
+      // Saturation checking
+      output_ = (output_ > saturation_) ? saturation_ : output_;
```
ethanmusser: This only handles positive saturation.

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,192 @@
...
+      saturation_ = saturation;
+    }
+
+    /**
+     * @brief Function to update the PID output.
+     * @param goal Final goal state for the controller.
+     * @param current Current state of the controller.
+     * @return Returns the Pid output
+     */
+    double Update(double goal, double current)
```
ethanmusser: Maybe provide the option to pass the time step in for the update computation.

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,192 @@
...
+
+    /**
+     * @brief Current derivative error found
+     */
+    double derivative_;
+
+    /**
+     * @brief Sumes up the error for calculating the I term
+     */
+    double i_error_;
```
ethanmusser, line 154: This member is never initialized.

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,192 @@
...
+
+  public:
+    /**
+     * @brief Construct a Pid object
+     * @param p_gain Proportional Gain
+     * @param d_gain Derivative Gain
+     * @param i_gain Integral Gain
+     * @param time_step Time difference between 2 updates
+     */
+    Pid(double p_gain, double d_gain, double i_gain, double time_step, double deadband, double saturation)
```
ethanmusser: Saturation and deadband should be optional parameters, many use-cases will not require these.

ethanmusser (changes requested):
See my comments below.

### Notes

- [ ] Fix your include guards.
- [ ] Fix the initialization structure for the `Pid` class.
- [ ] Give `Pid` a more descriptive name.
- [ ] Fix update loop saturation computation and code duplication
- [ ] Fix your formatting (and auto-formatter) to follow the [style guide](https://drake.mit.edu/styleguide/cppguide.html#Formatting).
   - [ ] Opening brackets follow on same line as [namespace](https://drake.mit.edu/styleguide/cppguide.html#Namespace_Formatting)/[class](https://drake.mit.edu/styleguide/cppguide.html#Class_Format)/[function](https://drake.mit.edu/styleguide/cppguide.html#Function_Declarations_and_Definitions) definition
   - [ ] [Namespace contents are not indented](https://drake.mit.edu/styleguide/cppguide.html#Namespace_Formatting)
   - [ ] [Hard line limit of 80 characters (except for enumerated exceptions)](https://drake.mit.edu/styleguide/cppguide.html#Line_Length)
   - [ ] [Use snake case for getters and setters](https://drake.mit.edu/styleguide/cppguide.html#Function_Names)
   - [ ] Indent class contents, giving `public`, `protected`, and `private` one-space indents
   - [ ] End all files with a newline

include/kodlab_mjbots_sdk/pid.h
chandravaran: I had changed it to this. I had known we could pass the variables we are getting directly. Also when I use the formatting option it makes it a line greater than 80. Any suggestions on fixing this?

include/kodlab_mjbots_sdk/pid.h
chandravaran: I was saving this in class variables, so I don't have to reinitialise the variable when I am running the program. It may be minimal but recreating a variable takes up some compute time.

chandravaran: Correct I will update it.

include/kodlab_mjbots_sdk/pid.h
chandravaran: Should I make saturation an array, then we could set the lower and upper limit of the saturation?

include/kodlab_mjbots_sdk/pid.h
chandravaran: When I try to pass the time_step variable, I am also trying to set the default to class variable time_step_. However, it is not a static variable so it is allowing me to set this. Then if I make it static then I can't initialise it like [here](https://drake.mit.edu/styleguide/cppguide.html#Function_Names). I would have to set it in the constructor.

include/kodlab_mjbots_sdk/pid.h
chandravaran: Have made them optional variables that can be set

include/kodlab_mjbots_sdk/pid.h
ethanmusser: I gave a solution in my above response...

As for formatting, put a line break between each initializer list element as described in [this style guide section](https://drake.mit.edu/styleguide/cppguide.html#Constructor_Initializer_Lists).

include/kodlab_mjbots_sdk/pid.h
ethanmusser: I think you are misunderstanding how the compiled code would execute here...

As written, every single computed value is saved as a data member.  As such, each computation step needs to be executed, and the result saved to memory (including values that will not be needed again, like `derivative_` and `output_`). This memory is allotted when the object is initialized and maintained while the object is kept in scope.

Alternately, you could write this function using temporary variables for any data that does not need to be kept.  For example, 
```cpp
float derivative = (error_ - d_error_) / time_step_;
```
When this code is compiled, the compiler will recognize that `derivative` is a temporary variable and (in most cases) optimize it out.  The resulting code will be more efficient at runtime. Although we are only talking about copying a couple of scalar values here, the principle is still important to understand.

include/kodlab_mjbots_sdk/pid.h
ethanmusser: It is unclear what you are asking here... 

I am simply saying that the current implementation only saturates positive output.  Any saturation functionality should cover both positive and negative outputs.

As written, with a saturation torque of `10 N-m`, 
| Ctrl. Torque | Output Torque |
|--------------|---------------|
|   5          |   5           |
|  15          |  10           |
| -15          | -15           |

include/kodlab_mjbots_sdk/pid.h
ethanmusser: No idea what you're asking here.  That link does not seem to be what you meant.

include/kodlab_mjbots_sdk/pid.h
chandravaran: Okay figured it out.

include/kodlab_mjbots_sdk/pid.h
chandravaran: Removed this

include/kodlab_mjbots_sdk/pid.h
chandravaran: Okay I get what you are saying. So I was thinking in a slightly different direction. So let us say we have a limit for a positive saturation as 10. This doesn't necessarily mean my lower bound would be -10. It could be some different value like 5. So I was thinking of making saturation an array that allows you to give 2 values, a lower bound and an upper bound. I have the implementation based on how I have explained it above. If you need me to change it to single saturation value let me know.

include/kodlab_mjbots_sdk/pid.h
chandravaran: My bad, I didn't double check the link I attached. 
Here is what I am taking about, the below picture is what you asked me to use for formatting. However, when I do this I get errors.
old_implementation

So implemented it like this.
code_implementation_current

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,180 @@
...
+ */
+
+#pragma once
+
+#include <math.h>
+
+#include <vector>
+
+#define HIGH_SATURATION_INDEX_ 0
+#define LOW_SATURATION_INDEX_ 0
```
ethanmusser, line 19: Constants should be named per the style guide (see [this section](https://drake.mit.edu/styleguide/cppguide.html#Constant_Names)).
```suggestion
#define kHighSaturationIndex 0
#define kHighSaturationIndex 0
```

ethanmusser, line 19: What is the point of these saturation indices?  The current method seems convoluted, why not simply store the limits in appropriately labelled members?

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,180 @@
...
+    PIDController(double p_gain, double d_gain, double i_gain, double time_step,
+      double deadband = 1.0,
+      std::vector<double> saturation = { INFINITY, -INFINITY })
+      : kp_(p_gain),
+      kd_(d_gain),
+      ki_(i_gain),
+      deadband_(deadband),
+      saturation_(saturation) {
+      time_step_ = time_step;
+    }
```
ethanmusser, line 44: Format per [this style guide section](https://drake.mit.edu/styleguide/cppguide.html#Constructor_Initializer_Lists).

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,180 @@
...
+
+#include <math.h>
+
+#include <vector>
+
+#define HIGH_SATURATION_INDEX_ 0
+#define LOW_SATURATION_INDEX_ 0
+
+namespace kodlab {
+
```
ethanmusser, line 22: Once again, namespace contents [should **NOT** be indented](https://drake.mit.edu/styleguide/cppguide.html#Namespace_Formatting).

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,180 @@
...
+      output_ = (output_ < saturation_[LOW_SATURATION_INDEX_])
+        ? saturation_[LOW_SATURATION_INDEX_]
+        : output_;
+      return output_;
+    }
+
+    /**
+     * @brief Destructor
+     */
+    ~PIDController() {}
```
ethanmusser, line 121: What's wrong with the default destructor?  

If this class is intended to be a base class, then this should be a virtual destructor to prevent a memory leak when destructing derived class instances (see explanation in [C.35](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-dtor-virtual); further discussion [here](https://en.cppreference.com/w/cpp/language/destructor)). 

Otherwise, I see no reason to overwrite the default destructor here.

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,180 @@
...
+#define HIGH_SATURATION_INDEX_ 0
+#define LOW_SATURATION_INDEX_ 0
+
+namespace kodlab {
+
+  /**
+   * @brief PIDController class that allows you to use PID control.
+   */
+  class PIDController {
+  public:
```
ethanmusser, line 27: Please review the style guide, this is incorrect formatting.  Proper formatting [described here](https://drake.mit.edu/styleguide/cppguide.html#Class_Format).

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,180 @@
...
+
+      // Checking the upper limit of the saturation level
+      output_ = (output_ > saturation_[HIGH_SATURATION_INDEX_])
+        ? saturation_[HIGH_SATURATION_INDEX_]
+        : output_;
+
+      // Checking the lower limit of the saturation level
+      output_ = (output_ < saturation_[LOW_SATURATION_INDEX_])
+        ? saturation_[LOW_SATURATION_INDEX_]
+        : output_;
```
ethanmusser, line 114: Replace this with a simple clamping function, either...
- use [`std::clamp` from the STL (part of the `algorithm` header)](https://en.cppreference.com/w/cpp/algorithm/clamp)
   ```cpp
         output_ = std::clamp(output_, saturation_low, saturation_high);
   ```
- use `std::min` and `std::max`
   ```cpp
         output_ = std::max(saturation_low, std::min(output_, saturation_high));
   ```
These are both much more readable.

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,180 @@
...
+
+    /**
+     * @brief Proportional Gain
+     */
+    double kp_;
+
+    /**
+     * @brief Derivative Gain
+     */
+    double kp_;
```
ethanmusser, line 137: `kp_` is defined twice, this should not compile.

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,180 @@
...
+
+    /**
+     * @brief Deadband of the system
+     */
+    double deadband_;
+
+    /**
+     * @brief Max control output
+     */
+    std::vector<double> saturation_{ INFINITY, -INFINITY };
```
ethanmusser, line 172: These absolutely should not be stored in a `std::vector`.  Since this is a fixed-length vector, you should be using `std::array`.  Also, the ordering of `{ max_sat, min_sat }` seems counterintuitive.

ethanmusser (commented):
Some more comments for you to consider. Some notes for you:
- You should really read through the entirety of the [style guide for this project](https://drake.mit.edu/styleguide/cppguide.html), a lot of the things pointed out below are addressed in this guide.
- `pid.h` has not been added to the `kodlab_mjbots_sdk` library in [`src/CMakeLists.txt`](https://github.com/KodlabPenn/kodlab_mjbots_sdk/blob/pid_class/src/CMakeLists.txt) in the remote repository. Unless you have this change added locally, your code here has not been run through the compiler.

include/kodlab_mjbots_sdk/pid.h
ethanmusser: This does not seem to be a formatting question, but rather an issue with your implementation.  I can't say for sure without a full code snippet.  But the initializer list (the initial approach) is the way that you should be doing this, as discussed above.

include/kodlab_mjbots_sdk/pid.h
ethanmusser: I agree that actuator saturation is not necessarily symmetric, and accepting upper and lower bounds makes sense.  But the implementation at the time of writing only clipped the upper end.

include/kodlab_mjbots_sdk/pid.h
chandravaran: Okay I understand now.

include/kodlab_mjbots_sdk/pid.h
chandravaran: Correct, I have implemented the array based implementation.

include/kodlab_mjbots_sdk/pid.h
chandravaran: Okay.

include/kodlab_mjbots_sdk/pid.h
```diff
@@ -0,0 +1,180 @@
+/**
+ * @file pid.h
```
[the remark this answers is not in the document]
chandravaran: Completed it.

include/kodlab_mjbots_sdk/pid.h
chandravaran: So I am planning on adding checks on the limits. So that when the user uses the code properly. I notice I made mistake here and didn't change the index value for the low saturation value. Also I defined these constant variables so the code could be more understandable. If the code just contained 0,1 a question may arise in the future as to what is in that index.

include/kodlab_mjbots_sdk/pid.h
chandravaran: Okay will change the functions

include/kodlab_mjbots_sdk/pid.h
ethanmusser: As long as the `saturation_` member documentation appropriately explains the indexing then this isn't totally necessary, but still good to have if you really want to be explicit.  

If you do want to add these constants, they should not be macros polluting the global namespace.  Instead, add them as namespaced constants within the class itself.  Since these values are fixed at compile time, they can be declared `constexpr`.  Further, they should be declared `inline` as described [here](https://stackoverflow.com/a/56950545).  

The benefit of doing it this way is that the constant is only accessible in the relevant namespace, and not everywhere that the header file is included.

include/kodlab_mjbots_sdk/pid.h
bosky2001: Done.

bosky2001:
1. Added Vector PID class. 
2. Implemented integral windup and PID class.
3. Addressed most style and implementation comments

- [ ]  Checking dead-band and integral windup implementation
- [ ] Adding scalability to Scalar PID class as-well
- [ ] checking styling

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+   * @param d_gain Derivative Gain
+   * @param i_gain Integral Gain
+   * @param time_step Time difference between 2 updates
+   */
+  PIDController(const VectorNS & p_gain,const VectorNS & d_gain,const VectorNS & i_gain, double time_step=0.001,
+    VectorNS deadband = 0.5*VectorNS::Ones(), VectorNS windup_limit = std::numeric_limits<Scalar>::max()*VectorNS::Ones(),
+    std::array<Scalar, 2> saturation = { -std::numeric_limits<Scalar>::max(), std::numeric_limits<Scalar>::max() })
+    : 
+      deadband_(deadband),
+      windup_limit_(windup_limit),
```
jdcaporale, line 45: Let's rename anything with the word windup to accumulator
and then in the comment describing the member variable, add a comment that says
"Used to prevent runaway integrator/windup"

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+      saturation_(saturation),
+      time_step_(time_step){
+
+    set_gains(p_gain,d_gain,i_gain); 
+    const int input_size = kp_.rows();
```
jdcaporale, line 50: This line is not being used. Should be removed.

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+
+public:
+  typedef Eigen::Matrix<Scalar,N,1> VectorNS;
+  
+  /**
+   * @brief Construct a vector Pid object
+   * @param p_gain Proportional Gain
+   * @param d_gain Derivative Gain
+   * @param i_gain Integral Gain
+   * @param time_step Time difference between 2 updates
```
jdcaporale, line 38: Add the additional parameters to the list in the comment

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+  /**
+   * @brief Construct a scalar Pid object
+   * @param p_gain Proportional Gain
+   * @param d_gain Derivative Gain
+   * @param i_gain Integral Gain
+   * @param time_step Time difference between 2 updates
+   */
+  PIDController(Scalar p_gain,Scalar  d_gain,Scalar  i_gain, double time_step=0.001,
+    VectorNS deadband = 0.5*VectorNS::Ones(),VectorNS windup_limit = std::numeric_limits<Scalar>::max()*VectorNS::Ones(),
+    std::array<Scalar, 2> saturation = { -std::numeric_limits<Scalar>::max(), std::numeric_limits<Scalar>::max() })
```
jdcaporale, line 67: Also put new lines/ fix spacing

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+    setpoint_ = VectorNS::Zero();
+    derivative_setpoint_ =VectorNS::Zero();    
+  }
+  
+  /**
+   * @brief Construct a scalar Pid object
+   * @param p_gain Proportional Gain
+   * @param d_gain Derivative Gain
+   * @param i_gain Integral Gain
+   * @param time_step Time difference between 2 updates
```
jdcaporale, line 63: Also missing params

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+  PIDController(Scalar p_gain,Scalar  d_gain,Scalar  i_gain, double time_step=0.001,
+    VectorNS deadband = 0.5*VectorNS::Ones(),VectorNS windup_limit = std::numeric_limits<Scalar>::max()*VectorNS::Ones(),
+    std::array<Scalar, 2> saturation = { -std::numeric_limits<Scalar>::max(), std::numeric_limits<Scalar>::max() })
+    : 
+      deadband_(deadband),
+      windup_limit_(windup_limit),
+      saturation_(saturation),
+      time_step_(time_step){
+
+    set_gains(p_gain,d_gain,i_gain);
```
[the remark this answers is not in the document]
jdcaporale:
```suggestion
    set_gains(p_gain, d_gain, i_gain); 
```

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+   * @brief Function to get deadband 
+   */
+  VectorNS get_deadband(){return deadband_;}
+
+
+  /**
+   * @brief Function to set the scalar Windup limit
+   * @param limit Windup limit
+   */
+  void set_windup(double limit){ windup_limit_ = limit*VectorNS::Ones();}
```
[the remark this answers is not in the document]
jdcaporale: And 
```suggestion
  void set_accumulation_limit(Scalar limit){ accumulation_limit_ = limit*VectorNS::Ones();}
```

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+  }
+
+  /**
+   * @brief Function to update setpoints and PID output.
+   * @param state Current state for the controller.
+   * @param derivative_state Current derivative state of the controller.
+   * @param time_step   Time difference between 2 updates
+   * @return Returns the Pid output
+   */
+  virtual VectorNS Update(const VectorNS & state,const VectorNS & derivative_state,double time_step) {
```
jdcaporale, line 263:
```suggestion
  virtual VectorNS Update(const VectorNS & state, const VectorNS & derivative_state, float time_step) {
```

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+    // Calculate the errors
+    // Proportional error
+    error_ = setpoint_ - state;
+    
+    // Derivative error
+    d_error_ =  derivative_setpoint_ - derivative_state;
+    
+    
+    // Calling UpdateWithError.
+    return UpdateWithError(error_, d_error_,time_step);;
```
jdcaporale, line 273:
```suggestion
    return UpdateWithError(error_, d_error_, time_step);
```

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+  }
+  /**
+   * @brief Function to update the PID output given the respective errors
+   * @param state Current state for the controller.
+   * @param derivative_state Current derivative state of the controller.  
+   * @return Returns the Pid output
+   */
+  VectorNS Update(const VectorNS & state,const VectorNS & derivative_state) {
+
+    return Update(state,derivative_state,time_step_);
```
jdcaporale, line 296:
```suggestion
    return Update(state, derivative_state, time_step_);
```

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+   * @brief Function to update setpoints and PID output.
+   * @param state Scalar state for the controller.
+   * @param derivative_state Scalar derivative state of the controller.
+   * @return Returns the Pid output
+   */
+  VectorNS Update(Scalar state,Scalar derivative_state) {
+    //checking if using for scalar PID
+    static_assert(N==1, "Using Scalar function in vector PID");
+    // converts scalar to vector form to calculate output
+    return Update(state*VectorNS::Ones(),derivative_state*VectorNS::Ones());
```
jdcaporale, line 286: Should have a time_step input I think ? 
And output a scalar
```suggestion
  Scalar Update(Scalar state, Scalar derivative_state) {
    //checking if using for scalar PID
    static_assert(N==1, "Using Scalar function in vector PID");
    // converts scalar to vector form to calculate output
    return Update(state*VectorNS::Ones(),derivative_state*VectorNS::Ones())(0);
```

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+   */
+  std::array<Scalar, 2> saturation_;
+
+  /**
+   * @brief Output of the function used internally
+   */
+  VectorNS output_ ;
+
+  static constexpr int kSaturationLowerLimitIndex = 0;
+  static constexpr int kSaturationUpperLimitIndex = 1;
```
jdcaporale, line 402: maybe add  the following replacing all the 0.001 default arguments
`static constexpr float kDefaultTimeStep = 0.001;`

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+   * @param d_gain Derivative Gain
+   * @param i_gain Integral Gain
+   * @param time_step Time difference between 2 updates
+   */
+  PIDCascade(Scalar p_gain,Scalar  d_gain,Scalar  i_gain, double time_step=0.001,
+    VectorNS deadband = 0.5*VectorNS::Ones(),VectorNS windup_limit = std::numeric_limits<Scalar>::max()*VectorNS::Ones(),
+    std::array<Scalar, 2> saturation = { -std::numeric_limits<Scalar>::max(), std::numeric_limits<Scalar>::max() })
+    : 
+      PIDController<Scalar, N>(p_gain,d_gain,i_gain,time_step,deadband,windup_limit,saturation){}
+
```
jdcaporale, line 445: Fix the comments (missing param and incorrect object name) and the spacing

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+  
+  VectorNS Update(Scalar state,Scalar derivative_state) {
```
jdcaporale, line 447:
```suggestion
  Scalar Update(Scalar state, Scalar derivative_state) {
```

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+    //checking if using for scalar PID
+    static_assert(N==1, "Using Scalar function in vector PID");
+    // converts scalar to vector form to calculate output
+    return Update(state*VectorNS::Ones(),derivative_state*VectorNS::Ones());
```
jdcaporale, line 451:
```suggestion
    return Update(state*VectorNS::Ones(), derivative_state*VectorNS::Ones())(0);
```

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+  }
+
+  VectorNS Update(const VectorNS & state,const VectorNS & derivative_state) {
+
+    return Update(state,derivative_state,time_step_);
+  }
```
jdcaporale, line 457:
```suggestion
  VectorNS Update(const VectorNS & state, const VectorNS & derivative_state) {

    return Update(state, derivative_state, time_step_);
  }  
```

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+    static_assert(N==1, "Using Scalar function in vector PID");
+    // converts scalar to vector form to calculate output
+    return Update(state*VectorNS::Ones(),derivative_state*VectorNS::Ones());
+  }
+
+  VectorNS Update(const VectorNS & state,const VectorNS & derivative_state) {
+
+    return Update(state,derivative_state,time_step_);
+  }  
+
```
jdcaporale, line 458: BELOW: Lots of spaces missing, change doubles to floats

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,533 @@
...
+   * @brief Stores the previous error for the d term
+   */
+  VectorNS v_error_ ;
+
+  /**
+   * @brief Stores the previous error for the d term
+   */
+  VectorNS v_cmd_ ;
+};
+
```
jdcaporale, line 530: Should add a VectorNS max_velocity_;  and accompanying setters and getters and than clamp v_cmd_ in the v_error_ calculation or the v_cmd_ we save.

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -306,9 +324,9 @@ class PIDController {
   virtual VectorNS UpdateWithError(const VectorNS & error_p, const VectorNS &  error_d, double time_step) {
     // Computing the integral error
     i_error_ += error_p*time_step;
-
-    //adding a limit on integral error to prevent windup
-    i_error_ = i_error_.cwiseMin(windup_limit_).cwiseMax(-windup_limit_);
+    
+    //adding a limit on integral error to prevent accumulator
```
jdcaporale, line 328:
```suggestion
    //adding a limit on integral error to prevent windup
```

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+  ~PIDController() {}
+
+
+protected:
+  
+  
+  /**
+   * @brief State Setpoint 
+   */
+  //Eigen::Matrix<Scalar,Eigen::Dynamic,1> setpoint_;
```
ShaneRozenLevy, line 380: Remove line

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+  VectorNS setpoint_;
+
+  /**
+   * @brief State derivative Setpoint
+   */
+  VectorNS derivative_setpoint_;
+  /**
```
ShaneRozenLevy, line 387: Consider passing through a formatter.

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+   * @brief Time difference between 2 updates
+   */
+  double time_step_;
```
ShaneRozenLevy, line 390: units?

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+   */
+  VectorNS kd_;
+
+  /**
+   * @brief Integral Gain
+   */
+  VectorNS ki_;
+
+  /**
+   * @brief Stores the previous error for the d term
```
ShaneRozenLevy, line 408: This comment doesn't seem to match the variable. I expect d error to be something like (vel-vel_error)

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+
+  /**
+   * @brief Sums up the error for calculating the I term
+   */
+  VectorNS i_error_ ;
+
+  /**
+   * @brief Error at each step.
+   */
+  VectorNS error_ ;
```
ShaneRozenLevy, line 420: Comment does not match what I suspect this variable is. I assume the variable is just the error for each state, rather than a timeseries as the comment suggests.

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+
+  /**
+   * @brief accumulator limit of the system if needed
+   */
+  VectorNS accumulator_limit_;
+  
+  /**
+   * @brief Max control output
+   */
+  std::array<VectorNS, 2> saturation_;
```
ShaneRozenLevy, line 435: why is this an array? Is this actually min and max? Make that clear in the comment.

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+
+};
+
+
+/**
+ * @brief PIDCascade class that allows you to use Cascade PID control.
+ */
+
+template <typename Scalar, int N=1>
+class PIDCascade: public PIDController<Scalar, N> {
```
ShaneRozenLevy, line 453: This should be in a separate file

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+#include <array>
+#include <vector>
+#include <algorithm>
+#include <Eigen/Dense>
+#include <tuple>
+
+namespace kodlab {
+
+/**
+ * @brief PIDController class that allows you to use PID control.
```
ShaneRozenLevy, line 24: this comment needs to be greatly expanded upon. This is not your basic pid controller. You will likely want to explain what its actually doing, especially with bells and whistles like deadband and accumulator limit.

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+ */
+
+template <typename Scalar, int N=1>
```
ShaneRozenLevy, line 27: Probably should add comments on what scalar and N are.

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+
+  /**
+   * @brief Deadband of the system
+   */
+  VectorNS deadband_;
+
+  /**
+   * @brief accumulator limit of the system if needed
+   */
+  VectorNS accumulator_limit_;
```
ShaneRozenLevy, line 430: Why is accumulator_limit_ only a max while saturation is likely a max and a min? These should probably be the same format.

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+                std::array<VectorNS, 2> saturation = 
+                    { -std::numeric_limits<Scalar>::max()*VectorNS::Ones(), 
+                       std::numeric_limits<Scalar>::max()*VectorNS::Ones() })
+    : 
+      deadband_(deadband),
+      accumulator_limit_(accumulator_limit),
+      saturation_(saturation),
+      time_step_(time_step){
+
+    set_gains(p_gain, i_gain, d_gain);
```
ShaneRozenLevy, line 60: All these variables probably should be initialized in the initializer list.  The setter functions are not actually doing anything.  Thoughts @jdcaporale

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+  /**
+   * @brief Function to get Kd gain 
+   */
+  VectorNS get_kd_gain(){return kd_;}
+  
+  /**
+   * @brief Function to set the scalar deadband
+   * @param deadband Deadband that needs to be set
+   */
+  void set_deadband(Scalar deadband) { deadband_ = deadband*VectorNS::Ones(); }
```
ShaneRozenLevy, line 145: Add the assert from line 120.

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+   * @brief Function to get deadband 
+   */
+  VectorNS get_deadband(){return deadband_;}
+
+
+  /**
+   * @brief Function to set the scalar accumulator limit
+   * @param limit accumulator limit
+   */
+  void set_accumulator(Scalar limit){ accumulator_limit_ = limit*VectorNS::Ones();}
```
ShaneRozenLevy, line 163: Add assert

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+  /**
+   * @brief Function to set the scalar  PID gains
+   * @param p_gain Proportional Gain
+   * @param d_gain Derivative Gain
+   * @param i_gain Integral Gain
+   */
+  
+  void set_gains(Scalar p_gain, Scalar i_gain, Scalar d_gain) {
+    //checking if using for scalar PID
+    static_assert(N==1, "Using Scalar function in vector PID");
```
ShaneRozenLevy, line 120: You should make this assert a function since you will be using it in multiple places.

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+    set_setpoints(setpoint*VectorNS::Ones(), 
+                  derivative_setpoint*VectorNS::Ones());
+  }
+  
+  /**
+   * @brief Function to get State setpoint
+   */
+  VectorNS get_state_setpoint(){return setpoint_;}
```
ShaneRozenLevy, line 232: Thoughts on using copy to return a potentially large vector? @jdcaporale This is used in a number of getters.

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+
+  /**
+   * @brief Function to get derivative state setpoint
+   */
+  VectorNS get_dstate_setpoint(){return derivative_setpoint_;}
+
+  std::pair<VectorNS, VectorNS> get_setpoints(){
```
ShaneRozenLevy, line 239: missing a comment

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+                const VectorNS & d_gain, 
+                double time_step = 0.001,
+                VectorNS deadband = 0.05*VectorNS::Ones(), 
+                VectorNS accumulator_limit = 
+                    std::numeric_limits<Scalar>::max()*VectorNS::Ones(),
+                std::array<VectorNS, 2> saturation = 
+                    { -std::numeric_limits<Scalar>::max()*VectorNS::Ones(), 
+                       std::numeric_limits<Scalar>::max()*VectorNS::Ones() })
+    : 
+      deadband_(deadband),
```
ShaneRozenLevy, line 55: Probably should assert Scalar is numeric if eigen does not: https://stackoverflow.com/a/26207551

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+#include <vector>
+#include <algorithm>
+#include <Eigen/Dense>
+#include <tuple>
+
+namespace kodlab {
+
+/**
+ * @brief PIDController class that allows you to use PID control.
+ */
```
ShaneRozenLevy, line 25: Your going to need to add a comment somewhere that the accumulator limit is a limit on integral error, not on ki * integral error. I would have expected it to be the later. @jdcaporale

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+    //adding a limit on integral error to prevent windup
+    i_error_ = i_error_.cwiseMin(accumulator_limit_)
+                       .cwiseMax(-accumulator_limit_);
+
+    // Finding the control output from the errors
+    output_ = kp_.array() * error_p.array() + kd_.array() * error_d.array() + 
+              ki_.array() * i_error_.array();
+
+    //Checking if in Deadband range
+    for(int i=0;i<error_p.rows();i++){
```
ShaneRozenLevy, line 357: Why not just use N?

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+      if ( std::abs(error_p.coeff(i)) < deadband_.coeff(i)) {
+          output_(i) =0; 
+          i_error_(i)=0;
+      } 
+    }
+    output_ = output_.cwiseMin(saturation_[1]).cwiseMax(saturation_[0]);
```
ShaneRozenLevy, line 363: Why add the `kSaturationLowerLimitIndex` if you don't use them

include/kodlab_mjbots_sdk/pid_controller.h
```diff
@@ -0,0 +1,648 @@
...
+   */
+  std::array<VectorNS, 2> saturation_;
+
+  /**
+   * @brief Output of the function used internally
+   */
+  VectorNS output_ ;
+
+  static constexpr int kSaturationLowerLimitIndex = 0;
+  static constexpr int kSaturationUpperLimitIndex = 1;
```
ShaneRozenLevy, line 443: I don't really like this design, since these variables are not available to the user. One option would be to create a class that is just a pair, but rename the functions for first and second to max and min and use that for saturation and accumulator. @jdcaporale

ShaneRozenLevy (changes requested):
I only took at a look at PID class and not cascade PID. Main blockers:

- Comments need a lot of work
- I'm not a fan of how `saturation_limit` is implemented.
- I think `accumulator_limit_` should also be a max and a min and it should probably be on ki * i_error

Once you fix these things and attempt to fix issues I might find in cascade PID, I'll take a look at cascade pid

include/kodlab_mjbots_sdk/pid_controller.h
jdcaporale: That makes sense to me, essentially limiting the effort that is allowed by the accumulator. That is more intuitive so let's do that.

include/kodlab_mjbots_sdk/pid_controller.h
jdcaporale: Agreed, if they can be in the initializer list they should be.

include/kodlab_mjbots_sdk/pid_controller.h
jdcaporale: Since this is a compile time check, the function may have requirements about how it is written. Look at asserts in Drake or Eigen or on google to check. My guess is a need for constexpr keyword .

include/kodlab_mjbots_sdk/pid_controller.h
jdcaporale: No need this isallowing us to set a equal deadband for every element.

include/kodlab_mjbots_sdk/pid_controller.h
jdcaporale: Not required.

include/kodlab_mjbots_sdk/pid_controller.h
jdcaporale: Agreed. Not trying to give access though, so make the output a  const reference.

include/kodlab_mjbots_sdk/pid_controller.h
ShaneRozenLevy: Then why do we have the assert for the pid gains?

include/kodlab_mjbots_sdk/pid_controller.h
jdcaporale: Can you explain a bit?

ShaneRozenLevy: sure something like. I think its important when having a the tuple of vectors for the torque min and max its important to know which is the min and max. Rather than relying on a private index only available to the class that uses it, lets make a class that has a max and a min using std::pair

```
template<VectorNS>
class Range{

Range(const & VectorNS min, const& VectorNS max):pair_(min,max)

const & VecotrNS min(){
return pair_.first
}

const & VecotrNS max(){
return pair_.second
}

private:
std::pair<VectorNS, VectorNS> pair_
}
```

include/kodlab_mjbots_sdk/pid_controller.h
bosky2001: Uhh, the saturation limits can also be uneven so I used min max but accumulator_limit is like a cutoff on i_error so I just used a max and used accumulator_limit_ as max and -accumulator_limit_ as min. I could do the same using the min max structure, its more probably would be more clear.

bosky2001:
Separated the classes into different files.

- [ ] Add comments on range implementation
- [ ]  Check documentation

include/kodlab_mjbots_sdk/pid_controller.h
jdcaporale: Added

include/kodlab_mjbots_sdk/pid_controller.h
jdcaporale: Added

include/kodlab_mjbots_sdk/pid_controller.h
jdcaporale: Still needs to be done

PR #75, opened 2022-10-25
Added joint torque measurments

ShaneRozenLevy:
Adds the ability to access moteus measured torque at the sdk level. This is helpful when using the moteus pd loop and could be used to detect cases where the moteus is running into the speed torque curve.


ethanmusser (approved):
This all looks good to me.

PR #77, opened 2022-11-17
Correct angular velocity units (now rad/s, not deg/s)

jdcaporale:
Changes the default units of the angular velocity in the kodlab_sdk from degrees per second. 
This puts the kodlab_sdk and the pi3hat defaults at odds, and so a few functions needed to be updated to convert the units. 

- Changed a constructor `IMUData(...)` and an `Update(...)` function
- Added constants for converting the units
- Changed the comments to match

Notes:
The way we would like to handle units in this repo requires that all units be in MKS and any variable or member not in MKS should have the units in the name, e.g. 
- `ang_vel` would be assumed to be in rad/s
- `ang_vel_dps` would be in degrees per second
- `x_velocity` would be assumed m/s
- `x_velocity_mm_s` then mm/s
It is ok for the MKS units to be added to the name, if the author feels it is good to clarify for whatever reason.

### DIFF
--- a/include/kodlab_mjbots_sdk/imu_data.h
+++ b/include/kodlab_mjbots_sdk/imu_data.h
@@ -50,7 +50,7 @@ private:
   Eigen::Quaternion<Scalar> quat_ = {1, 0, 0, 0};
 
   /**
-   * @brief IMUData euler angles in world frame
+   * @brief IMUData euler angles in world frame (default rad/s)
    * @note Always follow the default <b>Extrinsic X-Y-Z> convention
    */
   mutable kodlab::ValidatedCache<rotations::EulerAngles<Scalar>>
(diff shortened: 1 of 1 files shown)


include/kodlab_mjbots_sdk/imu_data.h
```diff
@@ -95,6 +95,16 @@ class IMUData
...
 
+  /**
+   * @brief Constant for taking degrees to radians
+   */
+  static constexpr float kDegsToRads_ = M_PI/180.0;
+  
+  /**
+   * @brief Constant for taking degrees to radians
+   */
+  static constexpr float kRadsToDegs_ = 180.0/M_PI;
```
ethanmusser, line 106: The correct way to handle these constants (if the definitions and implementation were separated into header and source files) would be with an `extern` definition ([see this post](https://stackoverflow.com/a/2043667)).  This is outside the scope of this PR, but it might be worth separating the definition and implementation of `IMUData` into separate files and using [explicit template specialization](https://isocpp.org/wiki/faq/templates#separate-template-fn-defn-from-decl).  Then modifying this constant definition to use `extern`.

ethanmusser (approved):
LGTM.  Small comment about defining constants to avoid pollution, but that can't be avoided with a class defined and implemented entirely in a header.

include/kodlab_mjbots_sdk/imu_data.h
jdcaporale: Just to be clear, this is not an action item for now, right?

PR #79, opened 2023-03-31
Loop validated cache

jdcaporale:
Adds a Loop-tied cache class and integrates into the control loop. 

Key changes:
- add virtual to `ValidatedCache` member functions for inheritance
- `LoopId` class, that houses static member we use to invalidate `ValidatedLoopCache` class instances
  - Housed in a `uint32_t`, which is large enough to avoid repeats offering unique loop ids
- `ValidatedLoopCache` class which auto invalidates when the control loop updates
- `MJBotsControlLoop` increments `LoopId` before every call of `Update()`

Design decisions / possible issues:
- `uint32_t last_loop_id_` and `LoopId loop_id_` members in the cache increases the memory overhead, especially for caching small data types, could be smaller if we use `uint8_t` but without the same uniqueness properties (at 1kHz, `uint8_t` would repeat 4 times a second)
- Implemented `operator=` overload to make using caches easier, but it does hide the validation from the user and could lead to mistakes

This was tested locally on hardware using print statements in a simple control loop.
```
class Test_Validated_Loop_Cache : public kodlab::mjbots::MjbotsControlLoop<ManyMotorLog> {
  using MjbotsControlLoop::MjbotsControlLoop;
  kodlab::ValidatedLoopCache<float> myValue;
  void Init() override {
    myValue = 0.5;
  }
  void Update() override {
    std::vector<float> torques(num_joints_, 0);
    robot_->SetTorques(torques);
    std::cout<<kodlab::LoopId::get()<<std::endl;
    if(!myValue.valid()){
        myValue = myValue.get() + 1;
    }
    // if(myValue.valid()){
    //     myValue = myValue.get() - 1;
    // }
    std::cout<<myValue<<std::endl;
  }
```

### DIFF
--- a/include/kodlab_mjbots_sdk/common_header.h
+++ b/include/kodlab_mjbots_sdk/common_header.h
@@ -94,6 +94,15 @@ public:
   }
 
   /**
+   * @brief Uses operator= to set the class data and marks it valid
+   * @param data valid data
+   */
+  void operator=(const T &data)
+  {
+    set(data);
+  }
+
+  /**
    * @brief Accessor for data (<b>does not check validity</b>)
    * @return data
    */
--- a/include/kodlab_mjbots_sdk/common_header.h
+++ b/include/kodlab_mjbots_sdk/common_header.h
@@ -44,7 +44,7 @@ template<typename T>
 class ValidatedCache
 {
 
-private:
+protected:
   /**
    * @brief Validation status of \c data_
    */
@@ -75,7 +75,7 @@ public:
    * @brief Returns the data status
    * @return \c true if data is valid, \c false otherwise
    */
-  bool valid() const { return valid_; }
+  virtual bool valid() const { return valid_; }
 
   /**
    * @brief Marks the class invalid so that future calls will change the cached
@@ -87,7 +87,7 @@ public:
    * @brief Sets the class data and marks it valid
    * @param data valid data
    */
-  void set(const T &data)
+  virtual void set(const T &data)
   {
     data_ = data;
     valid_ = true;
@@ -108,7 +108,7 @@ public:
    */
   T get()
   {
-    if (!valid_)
+    if (!valid())
     {
       std::cerr << "[WARN] Returning invalid data." << std::endl;
     }
--- a/include/kodlab_mjbots_sdk/loop_cache.h
+++ b/include/kodlab_mjbots_sdk/loop_cache.h
@@ -49,7 +49,7 @@ class LoopId
    * @note there can only be ONE loop using this type of cache at once to avoid 
    * prematurely invalidating caches, MAKE SURE THIS IS WHAT YOU WANT
    * 
-   * @return uint8_t 
+   * @return uint32_t 
    */
   static uint32_t increment(){
       return (++loop_id_);
@@ -76,6 +76,7 @@ class ValidatedLoopCache : public ValidatedCache<T>
   using ValidatedCache<T>::ValidatedCache;
   using ValidatedCache<T>::valid_;
   using ValidatedCache<T>::data_;
+  using ValidatedCache<T>::operator=;
 
   /**
    * @brief Returns the data status
@@ -95,7 +96,7 @@ class ValidatedLoopCache : public ValidatedCache<T>
   }
 
  protected:
-  uint8_t last_loop_id_; // loop id from the last set event, used to invalidate
+  uint32_t last_loop_id_ = -1; // loop id from the last set event, used to invalidate
   LoopId loop_id_; // Loop id, tied to control loop
 };
 
--- a/include/kodlab_mjbots_sdk/loop_cache.h
+++ b/include/kodlab_mjbots_sdk/loop_cache.h
@@ -1,0 +1,102 @@
+/**
+ * @file loop_cache.h
+ * @author J. Diego Caporale (jdcap@seas.upenn.edu)
+ * @brief Control loop validated cache and id for automatic invalidation synced 
+ * with loop
+ * @date 2023-03-30
+ * 
+ * @copyright Copyright (c) 2023 The Trustees of the University of Pennsylvania. 
+ * All Rights Reserved
+ * BSD 3-Clause License
+ */
+
+#pragma once
+#include <type_traits>
+#include <memory>
+
+#include "kodlab_mjbots_sdk/common_header.h"
+
+namespace kodlab{
+
+/**
+ * @brief Simple loop id counter class that allows the user to tell when the 
+ * control loop has looped. Useful for forcing recalculations and/or timing.
+ * 
+ */
+class LoopId
+{
+ public:
+  /**
+   * @brief Returns the current loop id
+   * 
+   * @return uint32_t 
+   */
+  static uint32_t get(){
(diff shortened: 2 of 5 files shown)


include/kodlab_mjbots_sdk/common_header.h
```diff
@@ -75,7 +75,7 @@ class ValidatedCache
    * @brief Returns the data status
```
jdcaporale, line 75: Need to add 
`virtual ~ValidatedCache(){};`

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -284,6 +287,7 @@ void MjbotsControlLoop<log_type, input_type, robot_type>::Run() {
     }
 
     // Calculate torques and log
+    kodlab::LoopId::increment(); // invalidate any caches that are tied to the loop by incrementing the id
```
ShaneRozenLevy, line 290: This is the correct place for this.

include/kodlab_mjbots_sdk/loop_cache.h
```diff
@@ -0,0 +1,103 @@
...
+#include <type_traits>
+#include <memory>
+
+#include "kodlab_mjbots_sdk/common_header.h"
+
+namespace kodlab{
+
+/**
+ * @brief Simple loop id counter class that allows the user to tell when the 
+ * control loop has looped. Useful for forcing recalculations and/or timing.
```
ShaneRozenLevy, line 23: Add a comment here about its thread safety or lack theirof.

include/kodlab_mjbots_sdk/common_header.h
```diff
@@ -87,19 +87,28 @@ class ValidatedCache
...
   /**
    * @brief Accessor for data (<b>does not check validity</b>)
    * @return data
    */
   T get()
   {
-    if (!valid_)
+    if (!valid())
     {
       std::cerr << "[WARN] Returning invalid data." << std::endl;
```
ShaneRozenLevy, line 113: Now that we are using this more, will this line terminate the code, or just print out an error. It might also be useful to provide a little more info about which cache it is. For now I believe we could print out T.

ShaneRozenLevy (approved):
Looks good. A few small things to discuss/document.

include/kodlab_mjbots_sdk/loop_cache.h
```diff
@@ -0,0 +1,103 @@
...
+      return (++loop_id_);
+  }
+
+ protected:
+  /**
+   * @brief Loop id shared between all instances acts a "global" loop id
+   * 
+   */
+  static uint32_t loop_id_; 
+};
```
ethanmusser, line 64: This is a mutable singleton with some possible issues.

First, philosophically, why use a singleton here?  It goes against the [ISO guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i3-avoid-singletons), and the resulting behavior could instead be maintained in a local scope.  E.g., a loop ID is maintained in the control loop and all `ValdatedLoopCache`s are constructed with a reference/pointer to that ID (and can use the same invalidation logic). 

That said, since this is a pretty high-level, integral thing, then a singleton could be fine.  But:
- this singleton could suffer from the [static initialization order fiasco](https://stackoverflow.com/a/35958423) if `LoopId` were to be referenced in another static variable initialization;
- `loop_id_` can be arbitrarily modified by children of `LoopId`;
- `LoopId` can be multiply constructed (which could be confusing).

You could alleviate some of these issues by making `LoopId` a [Meyer's](https://stackoverflow.com/a/1008289) [Singleton](https://laristra.github.io/flecsi/src/developer-guide/patterns/meyers_singleton.html).  I think something like this would work.

```cpp
class LoopId {
 public:
  static LoopId& get() {
    static LoopId loop_id;
    return loop_id;
  }

  static void set(uint32_t id) {
    auto loop_id = LoopId::get();
    loop_id = id;
  }

  static void increment() {
    auto loop_id = LoopId::get();
    loop_id++;
  }

  LoopId(const LoopId&) = delete;
  LoopId& operator=(const LoopId&) = delete;

private:
  LoopId() {}
  ~LoopId() {}

};
```

include/kodlab_mjbots_sdk/common_header.h
```diff
@@ -87,19 +93,28 @@ class ValidatedCache
...
 
+  /**
+   * @brief Uses operator= to set the class data and marks it valid
+   * @param data valid data
+   */
+  void operator=(const T &data)
+  {
+    set(data);
+  }
+
```
ethanmusser, line 110: I feel like this is fine despite the possible user confusion since the intended usage is to validate on new data and invalidate at a later point.

ethanmusser (commented):
Objections to `LoopId`, but looking good otherwise.

src/loop_cache.cpp
```diff
@@ -0,0 +1,13 @@
+/**
```
ShaneRozenLevy, line 1: What is the purpose of this file?

ShaneRozenLevy (approved):
LGTM, besides why loop_cache.cpp exists.

src/loop_cache.cpp
jdcaporale: Just placeholder making sure the new version is good. Could move implementation into it if that is something we want, but otherwise I will delete.

jdcaporale:
@ethanmusser Thoughts about this implementation? Took some of your advice, I think it should be ok now, and maybe even threadsafe with the addition of atomic.

src/loop_cache.cpp
ShaneRozenLevy: I'm always of the opinion that the implementation should live in the source file when possible.

jdcaporale: Since this is not templated, it's an easy move, so I will.

include/kodlab_mjbots_sdk/loop_cache.h
```diff
@@ -0,0 +1,139 @@
...
+  void set(const T &data) override
+  {
+    data_ = data;
+    valid_ = true;
+    last_loop_id_ = loop_id_.get();
+  }
+
+ protected:
+  uint32_t last_loop_id_ = -1; // loop id from the last set event, used to invalidate
+  LoopId loop_id_; // Loop id, tied to control loop
```
jdcaporale, line 136: Shouldn't have an instance

include/kodlab_mjbots_sdk/loop_cache.h
```diff
@@ -0,0 +1,139 @@
...
+  
+  /**
+   * @brief Sets the class data and marks it valid and saves loop id
+   * @param data valid data
+   */
+  void set(const T &data) override
+  {
+    data_ = data;
+    valid_ = true;
+    last_loop_id_ = loop_id_.get();
```
jdcaporale, line 131:
```suggestion
    last_loop_id_ = LoopId::get();
```

include/kodlab_mjbots_sdk/loop_cache.h
jdcaporale, line 136:
```suggestion
```

PR #67, opened 2022-09-16
Limb class

zacgong:
PR made in reference to #48 

Redid Limb Class without any merge errors or unnecessary files tracked in the commits/git.  Also should have fixed any formatting errors

Created a LimbBase class between JointBase and RobotBase

- Fully abstract base class with pure virtual kinematics and helper functions
- Incorporated into RobotBase with AddLimb function
- Simple_Limb example for 1DOF limb with joint at origin and straight end effector
- Testing in robot_example still a work in progress, having trouble compiling

Issue #48, opened 2022-08-06
Implement Limb Class to be used in RobotBase


jdcaporale:
Need a limb class `LimbBase` that can be used by RobotBase
First version:
- `Eigen:VectorXf ForwardKinematics( ... )`
- `Eigen:MatrixXf Jacobian( ... )`
- `Eigen:VectorXf InverseKinematics( ... )`
- `struct LimbConfig`
- `LimbConfig config`
- `std::vector<std::shared_ptr<JointBase>> joints`
- `virtual void Update()`
- helper functions and members 
- a clean 3DOF example

Future Versions:
- `std::vector<Eigen::Vector3f> jointFrames`
- Allow optimize of Jacobian and FK using jointFrames or upgrade to `ValidatedCache`
- General FK, Jac, IK Test Functions?

Far Future?:
- "upgrade" to DH parameters maybe
- URDF?
- anything else?

PR #81, opened 2023-04-06
Designing a LimbBase class 

jdcaporale:
Expanding on PR #67 based on Issue #48 ,

From there,

> 
> Created a LimbBase class between JointBase and RobotBase
> 
> -     Fully abstract base class with pure virtual kinematics and helper functions
> -     Incorporated into RobotBase with AddLimb function
> -     Simple_Limb example for 1DOF limb with joint at origin and straight end effector
> -     Testing in robot_example still a work in progress, having trouble compiling

What I added:

- Caching for kinematics and updated kinematic functions / logic. 
- Helper functions, e.g. new `joint_utils` namespace
- a 3DoF example, needs some work and testing

Still working out:
- What type of configuration is needed for limbs
    - Letting the user-made classes handle that might be beneficial
    - Also making a general RRR limb could be useful
- How could we handle multiple output options?
    - e.g. We often like having legs in polar and in cartesian, should the user make separate limbs with converting member/free functions or have a std::pair output or they add their own FKPolar function?

### DIFF
--- a/include/kodlab_mjbots_sdk/limb_base.h
+++ b/include/kodlab_mjbots_sdk/limb_base.h
@@ -28,7 +28,22 @@
 
 namespace kodlab
 {
-
+    /**
+     * @brief Simple base class for limbs with an arbitrary end effector space. 
+     * @details Includes virtual functions that can to be implemented by the 
+     * derived classes: 
+     *      ForwardKinematicsImpl(...) [req]
+     *      JacobianImpl(...) [req]
+     *      FKAndJacobianImpl(...) [optional, for optimization, MUST SET CACHES]
+     *      InverseKinematicsImpl(...) [optional]
+     *      EndEffectorVelocityImpl(...) [default: Jacobian * velocities]
+     * @note ForwardKinematicsImpl(...) and JacobianImpl(...) are required, but 
+     * are superceded in the Update() function by the implementation of 
+     * FKAndJacobianImpl(...) if it is non-empty, which is assumed to be 
+     * faster/better.   
+     * @tparam EndEffectorOutput, end effector space / output type of the 
+     * forward kinematics
+     */
     template<typename EndEffectorOutput = Eigen::VectorXf>
     class LimbBase {
       public:


include/kodlab_mjbots_sdk/limb_base.h
```diff
@@ -0,0 +1,224 @@
...
+        std::vector<Eigen::Matrix3f> actuator_orientations; 
+
+        /**
+         * @brief Matrix of zeros and ones for reordering joints in member 
+         * methods 
+         */
+        Eigen::MatrixXf selection_matrix;
+    };
+
+    template<typename EndEffectorOutput = Eigen::VectorXf>
```
jdcaporale, line 55: Missing documentation block

include/examples/quadruped_3dof_limb.h
```diff
@@ -0,0 +1,323 @@
+/**
+ * @file quadruped_3dof_limb.h
+ * @author Kodlab - J. Diego Caporale (jdcap@seas.upenn.edu)
+ * @brief A LimbBase derived class example of the common quadrupedal serial limb
+ * configuration with a point toe
```
ShaneRozenLevy, line 5: Provide some an example of a robot with this limb (mit mini cheetah)

include/examples/quadruped_3dof_limb.h
```diff
@@ -0,0 +1,323 @@
...
+    // Limb index is defined as follows
+    typedef uint8_t LimbIndex_t;
+    enum {
+        kFrontLeft = 0,
+        kRearLeft,
+        kFrontRight,
+        kRearRight
+    };
+
+    QuadrupedLimb(
```
ShaneRozenLevy, line 106: Provide a doxygen block for this comment.

include/examples/quadruped_3dof_limb.h
```diff
@@ -0,0 +1,323 @@
...
+        const std::vector<std::shared_ptr<kodlab::JointBase>> &joints,
+        const kodlab::LimbConfig &config,
```
ShaneRozenLevy, line 108: I thought we decided to nix this struct.

include/examples/quadruped_3dof_limb.h
```diff
@@ -0,0 +1,323 @@
...
+        LimbIndex_t limb_id = kFrontLeft)
+        : kodlab::LimbBase<Eigen::Vector3f>::LimbBase("", joints, config)
+    {
```
ShaneRozenLevy, line 111: You should run this code through an autofromatter. The brackets are wrong.

include/examples/quadruped_3dof_limb.h
```diff
@@ -0,0 +1,323 @@
...
+        is_front_limb_ = (int(limb_id) % 2) == 0;
+        is_left_limb_ = (int(limb_id) / 2) == 0;
+    }    
+    
+    Eigen::Vector3f ForwardKinematicsImpl(
+        std::vector<float>thetas) override 
+    {      
+        Eigen::Vector3f d0 = config_.actuator_offsets[0];
```
ShaneRozenLevy, line 119: The implementation here should be in a `.cpp`. The best way I found to due this with minimal overhead is by default when creating a class to create the source and header.

include/kodlab_mjbots_sdk/common_header.h
```diff
@@ -44,7 +44,7 @@ template<typename T>
 class ValidatedCache
 {
 
-private:
+protected:
```
ShaneRozenLevy, line 47: Im ignoring this class since I believed I already reviewed this pr.

include/kodlab_mjbots_sdk/limb_base.h
```diff
@@ -0,0 +1,224 @@
...
+#include "kodlab_mjbots_sdk/loop_cache.h"
+#include "kodlab_mjbots_sdk/joint_base.h"
+
+namespace kodlab
+{
+    /**
+     * @brief struct containing structural information to a limb
+     * @details all information provided by user at initialization
+     */
+    struct LimbConfig {
```
ShaneRozenLevy, line 35: I thought we were cutting this config since its too topology specific.

include/examples/quadruped_3dof_limb.h
```diff
@@ -0,0 +1,323 @@
...
+} //namespace example_helpers
+
+/**
+ * @brief A 3-DoF limb based on the common serial linkage in 12DoF quadrupeds
+ * @details Based on the limbs of popular quadrupeds
+ * @warning Needs to be tested, use at your own risk!!!!
+ * @todo    Needs to be tested ^^^
+ * @note    Needs to be tested ^^^
+ */
+class QuadrupedLimb : public kodlab::LimbBase<Eigen::Vector3f> {
```
ShaneRozenLevy, line 94: How will templating this to 3f play with the robot base using xf?

include/examples/simple_limb.h
```diff
@@ -0,0 +1,82 @@
...
+
+#include <Eigen/Dense>
+
+#include "kodlab_mjbots_sdk/limb_base.h"
+
+/**
+ * @brief A simple 1-DOF limb consisting of a single joint with a striaght
+ * link to an end-effector in Cartesian XYZ space
+ * @details A simple, single-degree-of-freedom limb implementation demonstrating 
+ * the specification of a Cartesian offset and implementation of kinematics
```
ShaneRozenLevy, line 23: Shouldn't mention cartesian offset since we are not specifying what that is.

include/examples/simple_limb.h
```diff
@@ -0,0 +1,82 @@
...
+ * methods
+ * @todo Implement running example
+ */
+class SimpleLimb : public kodlab::LimbBase<Eigen::Vector3f> {
```
ShaneRozenLevy, line 27: Shouldn't this be template to xf?

include/examples/quadruped_3dof_limb.h
```diff
@@ -0,0 +1,323 @@
...
+        toe_position = config_.actuator_offsets[3];
+        toe_jacobian.col(2) = knee_frame * yhat_skew * toe_position;
+        toe_position = config_.actuator_offsets[2] + R_CD * toe_position;
+        toe_jacobian.col(1) = hip_frame * yhat_skew * toe_position;
+        toe_position = config_.actuator_offsets[1] + R_BC * toe_position;
+        toe_jacobian.col(0) = abduc_frame * xhat_skew * toe_position;
+        toe_position = config_.actuator_offsets[0] + R_AB * toe_position;
+
+        // Set the cached kinematics
+        fk_.set(toe_position);
```
ShaneRozenLevy, line 279: Why are we expecting the user to set? Seems like bad design.

src/limb_base.cpp
```diff
@@ -0,0 +1,126 @@
...
+            jac_ = JacobianImpl(joint_positions_);
+        }
+        return jac_;
+    }
+
+    template<typename EndEffectorOutput>
+    std::vector<float> LimbBase<EndEffectorOutput>::InverseKinematics(
+                       const EndEffectorOutput &EE_pos)
+    {
+        return InverseKinematicsImpl(EE_pos);
```
ShaneRozenLevy, line 83: Should inverse kinematics output be cached?

ShaneRozenLevy (commented):
On the right track. A few key blockers:

- limbConfig struct
- Templating of example classes
- Lack of cache for ik
- Requiring users to set cache in the FkJacImpl function

include/examples/quadruped_3dof_limb.h
```diff
@@ -0,0 +1,323 @@
+/**
+ * @file quadruped_3dof_limb.h
+ * @author Kodlab - J. Diego Caporale (jdcap@seas.upenn.edu)
+ * @brief A LimbBase derived class example of the common quadrupedal serial limb
+ * configuration with a point toe
+ * @details 3-DoF limb implementation demonstrating the LimbBase class
+ * @date 2022-09-02
```
jdcaporale, line 7: wrong date

include/examples/quadruped_3dof_limb.h
jdcaporale: - [ ] Make working example

include/examples/quadruped_3dof_limb.h
jdcaporale: We did, changes were not finished and pushed, are now.

include/examples/quadruped_3dof_limb.h
jdcaporale: Before pushing, I think I want to force the changed files through one. Or maybe finally do the blameless change on the whole repo. Not sure.

include/examples/quadruped_3dof_limb.h
jdcaporale: I think it should work based on my understanding, but I don't know. I am rethinking having the Limbs vector at all and would love to figure that out, perhaps separate from this PR? i.e. remove it all together for now.

include/examples/simple_limb.h
jdcaporale: It is specific to the config for this file, I'll check that this is correct before merging

include/examples/quadruped_3dof_limb.h
jdcaporale: Could use a `std::pair<EndEffectorOutput, Eigen::MatrixXf>` output, or it could be hidden behind the `=operator` overload, but figured that this function is meant to be more advanced anyway.
The example is there to guide them through it, but I know it isn't super smooth.

src/limb_base.cpp
jdcaporale: Because the EE_pos is given by the user I didn't think it quite matched the pattern, happy to be overruled.

include/examples/quadruped_3dof_limb.h
jdcaporale: Also, I think I did this in a bunch of places. I've been reading too many other peoples' code and mentally switched convention.

jdcaporale: Actually I think the three of us have different conventions, which is worse, DEFINITELY IN NEED OF AN AUTOFORMATTER

include/examples/quadruped_3dof_limb.h
ShaneRozenLevy: I think thats fine.

include/examples/quadruped_3dof_limb.h
ShaneRozenLevy: I'd use the std::pair.

src/limb_base.cpp
ShaneRozenLevy: Hmm. I think you are correct. If the user calls IK multiple times with the same input thats on them.

ShaneRozenLevy: Or rather you are I think caching it sounds like a very bad idea.

include/examples/quadruped_3dof_limb.h
jdcaporale: Ok, assuming that is the case, two small questions exist.
first,medium: with my current implementation, I would need to also use std::optional i.e.
`
std::optional<std::pair<EndEffectorOutput, Eigen::MatrixXf>
`
or

`
std::pair<std::optional<EndEffectorOutput>, std::optional<Eigen::MatrixXf>>
`
second, minor : Can the optimizer remove that copy? can I ensure no copy?

ShaneRozenLevy: Whats the use case for returning `std::nullopt`?

For the minor question, here is a discussion. But the answer is the optimizer will likely remove the copy. https://stackoverflow.com/a/4643721

jdcaporale: The functionality as it currently in the `Update` function is to always run the FKAndJacobian first, and if the FK or Jac is not set, then run FKImpl and JacImpl. If `FKAndJacobianImpl` returns a pair, I would like to be able to tell that it ran the default implementation (which doesn't actually calculate).

examples/robot_example.cpp
```diff
@@ -66,7 +66,7 @@ int main(int argc, char **argv)
     std::vector<kodlab::mjbots::JointMoteus> joints;
     joints.emplace_back(100, 4, 1, 0,   1, 0);
     joints.emplace_back(101, 4,-1, 0, 5.0/3.0, 0);
-    // JointSharedVector<kodlab::mjbots::JointMoteus> joints;
+    // kodlab::joint_utils::JointSharedVector<kodlab::mjbots::JointMoteus> joints;
```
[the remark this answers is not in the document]
jdcaporale: Example of different patterns one can use to define the joints.

include/examples/quadruped_3dof_limb.h
ShaneRozenLevy: I think the `std::optional<std::pair<...>>` is the correct way to do it. Ideally there would be some way to programmatically check if the function was overridden but I don't think that is possible.

PR #87, opened 2023-06-07
Fix CTRL C Scoping

jdcaporale:
Addresses Issue #51. 
 
Moves the global variable into a single translation unit (common_header.cpp) and then accesses it via methods. 
Initial testing of this method worked, but this code remains untest and can be reviewed but DO NOT MERGE until tested. 

Also, handles a TODO that was stuck behind this fix, which moves LCM Subscriber source code into a cpp.


include/kodlab_mjbots_sdk/common_header.h
```diff
@@ -9,18 +19,25 @@
...
  */
-static std::atomic_bool CTRL_C_DETECTED(false);
+void ActivateCtrlC();
+
+/**
+ * @brief Returns the state of the CtrlC Flag
+ */
+bool CtrlCDetected();
+
+// Ctrl-C signal handling adapted from solo8 repo
```
ShaneRozenLevy, line 32: Does this comment still belong here?

ShaneRozenLevy (approved):
LGTM. One comment seemed out of place.

include/kodlab_mjbots_sdk/common_header.h
jdcaporale, line 32:
```suggestion
```

jdcaporale:
Test and ready.

PR #88, opened 2023-06-08
Soft start hot fix

jdcaporale:
The max torque setting is not being used.
Maybe the soft start shouldn't have the max torque in it at all ? 
As written it is broken. 

Possible paths forward:
- The solution that was implemented below.
- Make the TorqueLimiter::Constrain scalar function pass by reference like the vector version
- **Remove the max limit from the Soft Start entirely**, i.e. the pre-fix functionality (albeit with some unnecessary lines removed)  and torque limit in the joint torque function. This would also be more readable I think and is my preferred. 

Also, there are a couple of small changes:
- Fix a comment block that has the old structure
- Fix a constructor that has the old structure


ShaneRozenLevy (approved):
Solution lgtm.  unsure what your preferred solution is. My intuition for including a max torque in the soft start is to prevent a discontinuity in the torque limits ever.

jdcaporale:
> Solution lgtm. unsure what your preferred solution is. My intuition for including a max torque in the soft start is to prevent a discontinuity in the torque limits ever.

If we are using max_torque in both it shouldn't be discontinuous but I understand what you mean 
But from a naming perspective and readability it isn't great to hide that all in the soft-start

I will push this solution, so that things are working though.

Is discontinuous torque actually and issue? Maybe from a human safety perspective, since it would seem to lurch if it did happen.

ShaneRozenLevy:
Yeah its a safety concern. The whole point of the soft start is a continuous ramp in torque limit. How hard would it be to force the different torque limits to be the same?

ShaneRozenLevy:
Alternatively we could rename this to torque limiter and remove the robot wide torque limit?

jdcaporale:
Force them to what extent? In our implementation in the joints, they would be the same cause they are using the same member variable to do both. 
Could just change it to `TorqueConstrainWithSoftStart`, but maybe that is just semantics?

ShaneRozenLevy:
Or is there even another torque limiter in the system on master besides the one in the joints?

jdcaporale:
> Or is there even another torque limiter in the system on master besides the one in the joints?

I don't think there is. The functions are there for if people want to use them though.

jdcaporale:
Shane reviewed by looking over my shoulder.

Issue #80, opened 2023-03-31
JointBase needs virtual destructor


jdcaporale:
We use pointers to interact with JointMoteus and JointBase and so we should have a virtual destructor.

jdcaporale:
#89 fixes this

PR #89, opened 2023-06-17
Fix base classes, consts and virtual destructors

jdcaporale:
Fixes missing virtual destructors and the const correctness of many of the robot base member functions. 
A good opportunity to do more minor clean up on the base classes, if anything comes to mind.

This may require some const-correctness fixes to existing code.


jdcaporale:
Deals with #80

ShaneRozenLevy (approved):
Changes LGTM. What testing have you done?

ShaneRozenLevy:
I suspect if we do have memory leaks they are minor since we are not creating and removing a ton of these objects.

jdcaporale:
Tested and working on a quadrupedal robot, but not in the examples, best I can currently do.

PR #90, opened 2023-07-11
[RFC] Expanding Moteus Communication for Debug

lucienpeach:
Introduces new accessor functions within `joint_moteus.h` to allow for inclusion of additional registers [as defined](https://github.com/mjbots/moteus/blob/main/docs/reference.md#a2b-registers) in the moteus controller `reference.md`. Includes supporting architecture via `options` to allow user specification of desired registers to be read in controller definition. 

### **Change List**

- Provides new accessor functions to request additional registers from the moteus, as defined in the `QueryCommand` struct within `moteus_protocol.h`. These additional registers include:
  - `q_current`, the current in the Q phase measured in amperes.
  - `d_current`, the current in the D phase measured in amperes.
  - `temperature`, the current board temperature, measured in degrees Celsius.
  - `voltage`, the current input voltage.
  - `fault`, a fault code which will be set if the primary mode is 1 (Fault).
- Adds fault code enumeration (pulled from [fw/error.h](https://github.com/mjbots/moteus/blob/main/fw/error.h#L25) from the moteus repository) to enable reading of fault code information, if desired.
- Adds joint vector getter `GetJoints()` to `mjbots_hardware_interface.h` which returns a vector of shared pointers to objects of the `JointMoteus` class. This allows the new accessor functions in `joint_moteus.h` to be referenced from the controller level to allow for logging of the respective registers. These new accessor functions include:
  - `get_temperature()`, which provides the `temperature` register.
  - `get_q_current()`, which provides the `q_current` register.
  - `get_d_current()`, which provides the `d_current` register.
  - `get_voltage()`, which provides the `voltage` register.
  - `get_fault()`, which provides the `fault` register.
- Provides new boolean options within the `MjbotsHardwareInterface` class which dictate the request sent to the moteus by the pi3hat and the register packets to be sent from the moteus to the pi3hat. These new options include:
  -  `send_current`, which includes packets from `q_current` and `d_current` registers if true. 
  - `send_voltage`, which includes packets from `voltage` register if true.
  - `send_temperature`, which includes packets from `temperature` register if true.
  - `send_fault`, which includes packets from `fault` register if true. 

### **Upcoming Changes**

- Create a `joint_moteus.cpp` implementation file to clean up `joint_moteus.h` and prevent this file from becoming too convoluted.
- Streamline code to eliminate static resolution definition and reduce the number of options calls required. This would allow the user to specify the resolution of a register while simultaneously identifying that register to be included in packets from the moteus. 
  - This proposed change would require numerous type-related changes and edits within a number of low-level files, the specifics of which ought to be a topic of discussion and deliberation. 

### **Testing**

Hardware testing on the DOQ robot is planned to verify the effectiveness of this code in practice. This testing will primarily seek to validate logging of new registers from the moteus.

### **Questions**

There are a number of ways to build off this code and optimize it. Discussion of changes proposed above and suggestions on additional new changes are greatly appreciated.


include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -129,20 +157,43 @@ class JointMoteus: public JointBase
...
-              moteus_kd_(config.moteus_kd){}
+              moteus_kd_(config.moteus_kd),
+              temperature_(config.temperature),
+              q_current_(config.q_current),
+              d_current_(config.d_current),
+              voltage_(config.voltage) {}
         
         /**
-         * @brief Update the joint of the moteus. Converts rot/s to rad/s and saves mode
+         * @brief Update the joint of the moteus. Converts rot/s to rad/s and saves registers
```
jdcaporale, line 167: I would change this to state values 
```suggestion
         * @brief Update the joint of the moteus. Converts rot/s to rad/s and saves 
         other state values
```

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -129,20 +157,43 @@ class JointMoteus: public JointBase
...
          * 
          * @param reply_pos position reported by moteus [rot]
          * @param reply_vel velocity reported by moteus [rot/s]
          * @param reply_torque torque measured by the moteus [Nm]
          * @param mode      moteus mode
+         * @param reply_q_current current in the Q phase reported by moteus [A]
+         * @param reply_d_current current in the D phase reported by moteus [A]
+         * @param reply_voltage voltage reported by the moteus [V]
+         * @param reply_temperature temperature reported by the moteus [C]
+         * @param fault     fault code
```
jdcaporale, line 177:
```suggestion
         * @param fault fault code
```

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -129,20 +157,43 @@ class JointMoteus: public JointBase
...
+        void UpdateMoteus(
+            float reply_pos,
+            float reply_vel,
+            float reply_torque,
+            ::mjbots::moteus::Mode mode,
+            float reply_q_current,
+            float reply_d_current,
+            float reply_voltage,
+            float reply_temperature,
+            ::mjbots::moteus::Fault fault){
```
jdcaporale, line 188: Should we just take in the reply itself and parse some here?

src/mjbots_hardware_interface.cpp
```diff
@@ -43,6 +43,19 @@ void MjbotsHardwareInterface::InitializeCommand() {
...
+    }
+    if(send_voltage_){
+      cmd.query.voltage = ::mjbots::moteus::Resolution::kInt16;
+    }
+    if(send_temperature_){
+      cmd.query.temperature = ::mjbots::moteus::Resolution::kInt16;
+    }
+    if(send_fault_){
+      cmd.query.fault = ::mjbots::moteus::Resolution::kInt8;
+    }
```
jdcaporale, line 58: Should we always ask for this, or only when the mode returns a fault?

jdcaporale: e.g pseudo code

> 
> if( mode == FAULT and send_fault_) {
>   cmd.query.fault = kInt8
> } else{
>   cmd.query.fault = kIgnore}

src/mjbots_hardware_interface.cpp
```diff
@@ -70,7 +83,11 @@ MjbotsHardwareInterface::MjbotsHardwareInterface(std::vector<std::shared_ptr<Joi
                                                  std::optional<::mjbots::pi3hat::Euler> imu_world_offset_deg,
                                                  bool dry_run,
                                                  bool print_torques,
-                                                 bool send_pd_commands)
+                                                 bool send_pd_commands,
+                                                 bool send_current,
+                                                 bool send_voltage,
+                                                 bool send_temperature,
+                                                 bool send_fault)
```
jdcaporale, line 90: maybe change bool names to request_XXXX except for send_pd_commands, which is actually a set of values.

jdcaporale (commented):
High level comments: looks good!

- Update Moteus should take in a reply object
- maybe a good place for the use of std::optional? we should discuss
- A MessageRequestOptions sub structt might be nice to pass around
- I don't know if we need to constantly read fault, depending on how the mode message works

src/mjbots_hardware_interface.cpp
jdcaporale: depends on how the mode message works

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -33,6 +33,10 @@ struct MoteusJointConfig{
     float moteus_kp = 0;
     float moteus_kd = 0;
     float soft_start_duration_ms = 1;
+    float temperature = 0;
+    float q_current = 0;
```
ShaneRozenLevy, line 37: Why is a state in the config struct?

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -71,12 +79,20 @@ class JointMoteus: public JointBase
             float pos_max = std::numeric_limits<float>::infinity(),
             float soft_start_duration_ms = 1,
             float moteus_kp = 0,
-            float moteus_kd = 0)
+            float moteus_kd = 0,
+            float temperature = 0,
+            float q_current = 0,
```
ShaneRozenLevy, line 84: Why is a state being set by the constructor?

include/kodlab_mjbots_sdk/joint_moteus.h
ShaneRozenLevy: That really depends on how we are scoping, and I think both options are reasonable.

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -165,6 +216,41 @@ class JointMoteus: public JointBase
          */
         const ::mjbots::moteus::Mode & get_mode_reference()   const {return mode_; }
 
+        /**
+         * @brief Get the temperature register
+         * 
+         * @return float
+         */
+        virtual float get_temperature() const {return temperature_;}
```
ShaneRozenLevy, line 224: Why are these virtual?

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -41,6 +41,10 @@ struct ControlLoopOptions {
   bool dry_run = false;  ///< If true, torques sent to moteus boards will always be zero
   bool print_torques = false;  ///< If true, torque commands will be printed to console
   bool send_pd_commands = false; ///< If true, the control loop will send pd setpoints & gains in addition to ffwd torque commands
+  bool send_current = false;  ///< If true, the control loop will send d_current and q_current registers
```
ShaneRozenLevy, line 44: You should just have the resolution struct here instead of a bunch of bools.

ShaneRozenLevy (changes requested):
2 big issues are constructor for joint moteus, and the request data should be set via an option that is `QueryCommand` instead of the bools.

src/mjbots_hardware_interface.cpp
jdcaporale: After talking about it, we shouldn't change this

include/kodlab_mjbots_sdk/joint_moteus.h
jdcaporale: I misunderstood what we were talking about when we were talking earlier and now understand why you were so flabbergasted and agree.

lucienpeach:
Builds upon previous changes to allow requests to be sent via a `QueryCommand` struct that replaces the previous boolean method of requesting data while bolstering user optionality through resolution selection. Includes initial support for open-loop control and introduces safety checks via warning messages and `kodlab::ActivateCtrlC()` to alert the user of input resolution errors and to prevent unsafe and improper closed-loop control.

### **Change List**

- Introduces a source file `joint_moteus.cpp` to streamline header vs. implementation for the `JointMoteus` class. 
- Adds a library of preset `QueryCommand` structs at the `JointMoteus` class level for common combinations of requests and resolutions. This library is not comprehensive and thus the user can make adjustments at the controller level as desired. 
- Condensed `UpdateMoteus()` to accept a `reply_message` object of type `QueryResult`, as opposed to requiring a large number of individual reply fields (`reply_pos`, `reply_vel`, etc.) as inputs. This goes hand-in-hand with the previously stated change to allow for `QueryCommand` structs as user inputs.
- Because of the additional freedom provided to the user through the passing of a `QueryCommand` struct, it is now possible for a new and inexperienced user to encounter previously unseen issues. To prevent this from happening, and to inform the inexperienced user of these issues, a number of additions have been made:
  - Adds `LOG_WARN` messages within accessor functions at the `JointMoteus` class level which inform the user if the resolution of a requested data field is set to `kIgnore`. If the resolution is set to a different value, the accessor function provides the respective field at the resolution specified in the `QueryCommand` struct. These messages are of "warning" status and thus do not kill the control loop. 
  - Adds initial support for open-loop control via the introduction of a new option `open_loop` which is stored within the `options` struct. If this value is set to `true`, the control loop will not check to ensure that `position` and `velocity` resolutions are set to a value other than `kIgnore` (since these readings are not necessary for open-loop control). 
  - Introduces a new function in the `JointMoteus` class, `is_open_loop()`, which checks to ensure that the resolutions of `position` and `velocity` are set to values other than `kIgnore`. If either of these values are set to `kIgnore` in the `QueryCommand` input struct, `is_open_loop()` returns a boolean value of `true`, indicating that the current resolution configuration requires open-loop control. 
  - Outputs a `LOG_FATAL` message and runs `kodlab::ActivateCtrlC()` if `options.open_loop` is set to `false` but `is_open_loop()` outputs a value of `true`, indicating that open-loop control is necessary for the current resolution configuration but not yet specified. This effectively kills the control loop from within the constructor definition, preventing the user from running closed-loop control without reporting position or velocity. 
- Provides new accessor functions in the `JointMoteus` class:
  - `get_velocity()`, which overrides the respective virtual function in the `JointBase` class and enables `LOG_WARN` message support.
  - `get_position()`, which overrides the respective virtual function in the `JointBase` class and enables `LOG_WARN` message support.
  - `get_servo_torque()`, which overrides the respective virtual function in the `JointBase` class and enables `LOG_WARN` message support.
  - `get_measured_torque()`, which overrides the respective virtual function in the `JointBase` class and enables `LOG_WARN` message support.
  - `get_mode()`, which allows allows the user to directly access the mode (circumventing the current architecture, which requires accessing `GetJointModes()` via the `MjbotsHardwareInterface` class). Additionally provides support for `LOG_WARN` messages.
  - `get_query_command()`, which allows the user to access the current `QueryCommand` struct.
- Introduces the `all_unset()` function within `QueryResult` in the `moteus` namespace, which is used to determine if the fields within a given `QueryResult` struct are populated or are set to default values. Added a new enum value `kUnset` to both `Mode` and `Fault` enums within the `moteus` namespace to represent these unset/default values (set to -1).
  - In conjunction with the `any_set()` function which exists within `QueryCommand` in the `moteus` namespace, these additions are intended to provide more accurate insight into potential drops in CAN communication via terminal messages.
- Minor changes to settings and naming conventions. 
  - Changed `send_pd_commands` bool to `use_pd_commands` to better encompass the utility of this option.
  - Adjusted maximum commanded torque resolution `res.maximum_torque` to `kInt16` resolution from `kInt8` to prevent cutoffs of 0.5 Nm when `use_pd_commands` is set to `true`.

### **Testing**
Testing on DOQ has confirmed proper communication with the robot and functionality of safety features in both open- and closed-loop control schemes (toggled via the `open_loop` option). Plotting has demonstrated reasonable values for all requested fields and control loop communications function without packet loss at 1000 Hz even with a fully comprehensive `QueryCommand` struct that returns all possible data fields. 

### **Questions** 
Open to suggestions on `QueryCommand` library changes or additions. Interested in thoughts of more robust open-loop support in future PRs.

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -1,7 +1,7 @@
 /**
  * @file joint_moteus.h
  * @author J. Diego Caporale (jdcap@seas.upenn.edu)
- * @brief Moteus powered joint class
+ * @brief Moteus powered joint class header
```
ShaneRozenLevy, line 4: You don't need to change this. The header is for the class.

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -43,6 +31,30 @@ struct MoteusJointConfig{
 class JointMoteus: public JointBase
 {
     public:
+        /**
+         * @brief Define a default resolution struct
+         *
```
ShaneRozenLevy, line 36: Fix the extra line here.

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -43,6 +31,30 @@ struct MoteusJointConfig{
 class JointMoteus: public JointBase
 {
     public:
+        /**
+         * @brief Define a default resolution struct
+         * 
+         */
+        static const ::mjbots::moteus::QueryCommand kDefaultQuery;
```
ShaneRozenLevy, line 38: @jdcaporale Do you think this is the correct place for these to live? Sort of torn if they should live here or in a namespace.

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -43,6 +31,30 @@ struct MoteusJointConfig{
...
+
+        /**
+         * @brief Define a resolution struct that includes torques
+         * 
+         */
+        static const ::mjbots::moteus::QueryCommand kTorqueQuery;
+
+        /**
+         * @brief Define a debug resolution struct
```
ShaneRozenLevy, line 47: What is the difference between `kDebugQuery` and `kComprehensiveQuery`? Its not obvious from the comments.

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -103,129 +114,193 @@ class JointMoteus: public JointBase
...
+
         /**
          * @brief Destroy the JointMoteus object, virtual to ensure proper 
          * destruction of derived classes
          */
         virtual ~JointMoteus(){}
 
         /**
-         * @brief Update the joint of the moteus. Converts rot/s to rad/s and saves mode
+         * @brief Determines whether open loop options should be enabled
```
ShaneRozenLevy, line 143: Is this function calculating if open loop, or just returning a bool? If its calculating, what determines if the moteus is open loop? Make that clear in the comment

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -103,129 +114,193 @@ class JointMoteus: public JointBase
...
          */
-        int get_can_bus() const {return can_bus_;}
+        int get_can_bus() const;
+
+        /**
+         * @brief Get the QueryCommand object
+         * 
+         * @return const ::mjbots::moteus::QueryCommand
+         */
+        const ::mjbots::moteus::QueryCommand get_query_command() const;
```
ShaneRozenLevy, line 168: Either return a const reference or return by value. Don't return by const value.

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -103,129 +114,193 @@ class JointMoteus: public JointBase
...
          * 
          * @return const ::mjbots::moteus::Mode& 
          */
-        const ::mjbots::moteus::Mode & get_mode_reference()   const {return mode_; }
+        const ::mjbots::moteus::Mode & get_mode_reference() const;
+
+        /**
+         * @brief Get the mode 
+         * 
+         * @return const ::mjbots::moteus::Mode
```
ShaneRozenLevy, line 180: These ai generated comments are not useful. Either make them useful or remove them.

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -103,129 +114,193 @@ class JointMoteus: public JointBase
...
+         */
+        const ::mjbots::moteus::Mode get_mode() const;
```
ShaneRozenLevy, line 182: These should probably be [[nodiscard]]

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -103,129 +114,193 @@ class JointMoteus: public JointBase
...
+    float zero_offset = 0;
+    float gear_ratio = 1.0;
+    float max_torque = std::numeric_limits<float>::infinity();
+    float pos_min = -std::numeric_limits<float>::infinity();
+    float pos_max = std::numeric_limits<float>::infinity();
+    float moteus_kp = 0;
+    float moteus_kd = 0;
+    float soft_start_duration_ms = 1;
 };
+
```
ShaneRozenLevy, line 304: Why white space change?

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -190,14 +191,23 @@ MjbotsControlLoop<log_type, input_type, robot_type>::MjbotsControlLoop(std::shar
     joints_moteus.push_back(
         std::dynamic_pointer_cast<kodlab::mjbots::JointMoteus>(j));
   }
+  // Open loop check
+  for(const auto & j : joints_moteus){
+      if(!options.open_loop && j->is_open_loop()){
+        LOG_FATAL("Open loop option not set, not receiving position and/or velocity");
```
ShaneRozenLevy, line 197: Lets make this a little more clear. Maybe:

`ControlLoopOptions has open_loop=false, but joint moteus QueryCommand struct does not query position and or velocity data. `

include/kodlab_mjbots_sdk/mjbots_control_loop.h
```diff
@@ -190,14 +191,23 @@ MjbotsControlLoop<log_type, input_type, robot_type>::MjbotsControlLoop(std::shar
...
   // Initialize mjbots_interface
   mjbots_interface_ = std::make_shared<kodlab::mjbots::MjbotsHardwareInterface>(
       std::move(joints_moteus), options.realtime_params,
       options.imu_mounting_deg, options.attitude_rate_hz,
       robot_->GetIMUDataSharedPtr(), options.imu_world_offset_deg,
       options.dry_run,
       options.print_torques,
-      options.send_pd_commands);
+      options.use_pd_commands,
+      options.open_loop
```
ShaneRozenLevy, line 209: Why does the `MjbotsHardwareInterface` take in the open_loop bool since it is not used in the class?

src/joint_moteus.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+
+#include "kodlab_mjbots_sdk/joint_moteus.h"
+
+namespace kodlab{
+namespace mjbots{
+
+// kDefaultQuery defintion
+const ::mjbots::moteus::QueryCommand JointMoteus::kDefaultQuery{
+  ::mjbots::moteus::Resolution::kInt8,
```
ShaneRozenLevy, line 20: I feel like we should comment this so its readable to the user.

src/joint_moteus.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+// kComprehensiveQuery definition
+const ::mjbots::moteus::QueryCommand JointMoteus::kComprehensiveQuery{
+  ::mjbots::moteus::Resolution::kInt8,
+  ::mjbots::moteus::Resolution::kInt16,
+  ::mjbots::moteus::Resolution::kInt16,
+  ::mjbots::moteus::Resolution::kInt32,
+  ::mjbots::moteus::Resolution::kInt16,
+  ::mjbots::moteus::Resolution::kInt16,
+  ::mjbots::moteus::Resolution::kIgnore,
```
ShaneRozenLevy, line 68: what is kIgnore in comprehensive? That doesn't make sense.

src/joint_moteus.cpp
ShaneRozenLevy: Once these are commented I will want to give this a look over with Diego to make sure we are happy with the standards.

src/joint_moteus.cpp
```diff
@@ -0,0 +1,294 @@
...
+                    mode_ = reply_message.mode;
+                    q_current_ = reply_message.q_current;
+                    d_current_ = reply_message.d_current;
+                    voltage_ = reply_message.voltage;
+                    temperature_ = reply_message.temperature;
+                    fault_ = reply_message.fault;
+}
+
+bool JointMoteus::is_open_loop() {
+  bool rv;
```
ShaneRozenLevy, line 139: I know I initially gave you the way to use this function, but right now its super silly. Consider instead:

```
return (query_type_.position == ::mjbots::moteus::Resolution::kIgnore ||
  query_type_.velocity == ::mjbots::moteus::Resolution::kIgnore);
```

src/joint_moteus.cpp
```diff
@@ -0,0 +1,294 @@
...
+const ::mjbots::moteus::Mode & JointMoteus::get_mode_reference() const {
+  if(query_type_.mode == ::mjbots::moteus::Resolution::kIgnore){
+    LOG_WARN("Mode resolution is set to kIgnore, while attempting to log register");
+  }
+  return mode_;
+}
+
+const ::mjbots::moteus::Mode JointMoteus::get_mode() const {
+  if(query_type_.mode == ::mjbots::moteus::Resolution::kIgnore){
+    LOG_WARN("Mode resolution is set to kIgnore, while attempting to log register");
```
ShaneRozenLevy, line 170: Comment should not mention log. Instead maybe say "get value".

ShaneRozenLevy, line 170: Should also mention this is in the `QueryCommand`,  resolution is also set in the data sent to the moteus.

ShaneRozenLevy (changes requested):
Much closer than last time. Comments are mostly just clean up and making everything super duper clear.

src/joint_moteus.cpp
lucienpeach: The value associated with this kIgnore is "rezero_state", for which there is no documentation within the mjbots `reference.md` file. As such, I opted to set the resolution to kIgnore, as I was unsure what information this contains / how it would interact with the closed-loop control (if at all). 

Think that it would be better to include all fields, like suggested, but either want to know what this field does or perhaps change the name of the query.

ShaneRozenLevy: I think thats fine, just add a comment.

include/kodlab_mjbots_sdk/joint_moteus.h
lucienpeach: Followed formatting used in `moteus_protocol.h` and `joint_base.h`, but notice now that other files (such as `robot_base.h`) do not have whitespace here. Google Style guide on [namespace formatting](https://google.github.io/styleguide/cppguide.html#Namespace_Formatting) seems to suggest a line of vertical whitespace in this context, although the style guide also recommends limiting vertical whitespace when possible. 

Fine with either decision.

ShaneRozenLevy: Either is fine. Generally isolated white space changes are frowned upon.

include/kodlab_mjbots_sdk/joint_moteus.h
lucienpeach: Changed comment and function name to better describe the utility of this function. Open to additional changes/suggestions.

include/kodlab_mjbots_sdk/joint_moteus.h
lucienpeach: Relocated to the bottom of the file for the time being, in keeping with the comparable change to source file.

src/joint_moteus.cpp
```diff
@@ -0,0 +1,305 @@
...
+}
+
+float JointMoteus::get_servo_torque() const {
+  if(query_type_.torque == ::mjbots::moteus::Resolution::kIgnore){
+    LOG_WARN("Torque resolution in QueryCommand is set to kIgnore, while attempting to get value");
+    return 0;
+  } else {
+    return servo_torque_;
+  }
+}
```
griffinaddison, line 138: this whole function shouldn't be here because query torque only handles measured torque. parent class joint_base already has a get servo torque function.
```suggestion
```

include/kodlab_mjbots_sdk/joint_moteus.h
```diff
@@ -103,129 +90,219 @@ class JointMoteus: public JointBase
...
+         * @return float
+         */
+        float get_velocity() const override;
+
+        /**
+         * @brief Get the torque command. Overrides function in joint_base.h
+         * 
+         * @return float
+         */
+        float get_servo_torque() const override;
```
griffinaddison, line 183: See comment in cpp. get servo torque already defined in parent class, was throwing resolution warning when not part of query.

```suggestion
```

src/joint_moteus.cpp
```diff
@@ -0,0 +1,305 @@
...
+
```
griffinaddison, line 139:
```suggestion
```

jdcaporale (approved):
Small change added, but otherwise this has been a long time coming. Great job!

include/kodlab_mjbots_sdk/joint_moteus.h
jdcaporale: I think this is fine.

